PR #30, opened 2021-07-22
doc: Scaling Use Cases

msbrogli:


text/0000-scaling-use-cases.md
```diff
@@ -0,0 +1,115 @@
...
+            5 | Dave
+            6 | Eve
+            7 | Alice
+```
+
+Notice that each index belongs to only one user. But the same user may have one or more indexes.
+
+This approach does not scale easily because the global wallet will have to keep track of more and more addresses over time. The more users, the more addresses. The more addresses, the more memory is required.
+
+Another challenge is to scan the transactions of the global wallet. The most common stop criteria for the scanner is the GAP Limit, which stops the scanner when `N` addresses in a row are empty. An address is empty when it has no transactions.
```
pedroferreira1, line 52: Given that it's a generic doc for use cases, I think we can explain when the gap limit problem can happen. I though about starting this paragraph like this:

Depending on the moment that the use case associates a new address to a user, we may face another challenge to know when to stop scanning the wallet address. E.g. an exchange associates the address to a user as soon as the user requests it but he might never send a transaction to it, on the other hand, some use cases can wait until the moment the user needs to send its first transaction before associating an address to him. In this scenario, no addresses would be unused.

text/0000-scaling-use-cases.md
```diff
@@ -0,0 +1,115 @@
...
+
+A Full-Length Wallet Service is a service that keeps track of all addresses in the blockchain and allows the creation of wallets.
+
+Wallets can have two types: manual or xpub.
+
+- Manual wallets have a list of addresses that is manually maintened, i.e., one must manually add or remove addresses from the wallet.
+- Xpub wallets have an xpub, and the addresses are scanned using the GAP Limit stop criteria.
+
+In this case, it is easier to use xpub wallets and ensure that the deposits meets the requirements of the GAP Limit.
+
```
pedroferreira1, line 109: We could add a section explaining a bit of how HD Wallets work (BIP32 and BIP44)

text/0000-scaling-use-cases.md
```diff
@@ -0,0 +1,115 @@
...
+The challenges here are the same regardless the way the wallet is associated with a user.
+
+The naive implementation here is to run one wallet per user. But it would consume a lot of resources (CPU and memory) and it is hard to scale to millions of users.
+
+A better implementation is to run a pool of wallets that automatically starts and stops wallets on demand. For instance, when a request is made to a wallet that is not running, the pool starts that wallet automatically. When a wallet is idle for too long, it is automatically stopped.
+
+Such pool would consume resources according to the number of concurrent users (instead of total users). It can easily scale up through the addition of more servers to the pool.
+
+## Full-length Wallet Service
```
andreabadesso, line 99: I think we should add a note explaining that this option is not currently available as it is being tested internally

andreabadesso (changes requested):

text/0000-scaling-use-cases.md
```diff
@@ -0,0 +1,115 @@
...
+
+This approach has a similar scaling issue of the Global Wallet. The wallet will have to keep track of all these derivation paths and their addresses.
+
+Another challenge is to scan the derivation paths. The most common stop criteria for the scanner is the GAP Limit M/N, which stops the scanner when `M` derivation paths in a row are empty. It considers that a given derivation path is empty when its first `N` addresses are empty.
+
+## One Wallet Per User
+
+This case differs from the previous ones because there will be multiple wallets running.
+
+First of all, there are multiple ways to associate a wallet with a user. To illustrate, some ways are: (i) one seed per user, (ii) one global seed with one passphrase per user, or (iii) one global seed with one derivation path per user.
```
obiyankenobi, line 89: Isn't option (iii) the same as the last section (Global Wallet with Derivation Paths)?

PR #31, opened 2021-07-22
High level of Nano Contracts

msbrogli:


text/0000-nano-contracts.md
```diff
@@ -0,0 +1,431 @@
...
+├─────────────────────┬──────────────────────────────────────┤
+│       Inputs        │               Outputs                │
+├─────────────────────┼──────────────────────────────────────┤
+│  2,500 HTR          │  2,500 HTR                           │
+│  Nano Contract #1   │  P2PKH                               │
+│  collect_prize()    │  HBpGrjZcYNVJrZTy4Yct7HtUdLM1ToCPi3  │
+└─────────────────────┴──────────────────────────────────────┘
+</pre>
+
+Finally, a simple call to a method is similar to a deposit or withdrawal but with amount zero.
```
obiyankenobi, line 186: Will we have methods that can be only used in the tx inputs and others in the outputs? Should it be explicit in the blueprint?

`collect_prize` is a method just like `make_a_bet` or any other method that can be called with amount 0. How can we know where they are allowed? Does it make sense to have a method with amount 0 called in the inputs?

text/0000-nano-contracts.md
```diff
@@ -0,0 +1,431 @@
...
+	def make_a_bet(self, address: Address, amount: Amount, score: str) -> None:
+		"""Make a bet."""
+		if timestamp > date_last_offer:
+			self.fail('cannot place bets after 2021-01-01 17:00')
+		self.bets[score].add_bet(address, amount)
+		self.total += amount
+
+	@public
+	def set_result(self, result: str) -> None:
+		"""Set result. Used by the oracle."""
```
obiyankenobi, line 267: Only the oracle can call this? Would be interesting if anyone can call it, as long as the message is properly signed by the oracle.

PR #32, opened 2021-08-02
feat: NFT standard

pedroferreira1:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/feat/nft-standard/text/0000-nft-standard.md)

### DIFF
--- a/text/0000-nft-standard.md
+++ b/text/0000-nft-standard.md
@@ -7,27 +7,27 @@
 # Summary
 [summary]: #summary
 
-This document presents a standard to be followed when creating a NFT transaction and its metadata. The idea behind having a standard is to be able to identify NFT tokens and to be easy for any plataform that would like to integrate Hathor NFTs.
+This document presents a standard to be followed when creating an NFT transaction and its metadata. The idea behind having a standard is to be able to identify NFT tokens and to be easy for any plataform that would like to integrate Hathor NFTs.
 
 # Motivation
 [motivation]: #motivation
 
-Creating a NFT standard is important mainly for 3 reasons:
+Creating an NFT standard is important mainly for 3 reasons:
 
-1. There are some requirements that a NFT token must fulfill in order to have its digital asset shown in our explorer, so it's important that they are described here.
+1. There are some requirements that an NFT token must fulfill in order to have its digital asset shown in our explorer, so it's important that they are described here.
 1. Having a similar structure in most NFTs (also for its metadatas) facilitates the integration with any platform to list and show information about NFTs created on Hathor. 
-1. Identifying a NFT is important to show specific information in the explorer and wallets.
+1. Identifying an NFT is important to show specific information in the explorer and wallets.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
 Non-fungible tokens (NFTs) are unique, indivisible digital assets created on blockchains. HTR token, for instance, is a fungible token and 1 HTR of mine can be exchanged with 1 HTR of anyone without difference because all tokens are equal. Each NFT’s uniqueness can be proven by a unique identifier.
 
-In Hathor Network a NFT is a new custom token created and its unique identifier is the transaction hash.
+In Hathor Network an NFT is a new custom token created and its unique identifier is the transaction hash.
 
 ## Transaction standard
 
-A transaction can be identified as a NFT creation if it has the following structure:
+A transaction can be identified as an NFT creation if it has the following structure:
 
 1. Transaction version is 2 (TOKEN_CREATION_TRANSACTION).
 1. All inputs are HTR inputs.
@@ -37,11 +37,11 @@ A transaction can be identified as a NFT creation if it has the following struct
 1. We have N outputs with custom token created value (the value can be any number).
 1. We can have 1 or 0 HTR output with change value.
 
-Any transaction that has this structure will be identified as a NFT and the explorer/wallet screens will show specific NFT information for them.
+Any transaction that has this structure will be identified as an NFT and the explorer/wallet screens will show specific NFT information for them.
 
 ## Output script data
 
-The output script data is the most important piece of a NFT transaction because it represents the data that is being uniquely identified in the blockchain.
+The output script data is the most important piece of an NFT transaction because it represents the data that is being uniquely identified in the blockchain.
 
 The data can be any string, e.g. '#A123' as a serial number, or 'https://mywebsite.com/', or even 'ipfs://ipfs/<hash>/filename'.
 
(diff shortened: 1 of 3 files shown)


text/0000-nft-standard.md
```diff
@@ -0,0 +1,105 @@
...
+
+Any transaction that has this structure will be identified as a NFT and the explorer/wallet screens will show specific NFT information for them.
+
+## Output script data
+
+The output script data is the most important piece of a NFT transaction because it represents the data that is being uniquely identified in the blockchain.
+
+The data can be any string, e.g. '#A123' as a serial number, or 'https://mywebsite.com/', or even 'ipfs://ipfs/<hash>/filename'.
+
+The most common use case is to represent a digital asset, then the NFT data usually redirects to an image/video/audio file. In that case, in order for us to add the digital asset in our explorer, it's required that this URL uses a immutable protocol, e.g. [IPFS](https://en.wikipedia.org/wiki/InterPlanetary_File_System). **Important to notice that your IPFS URL to be considered immutable must be like ipfs://ipfs/<hash>/filename and not https://ipfs.io/ipfs/<hash>/filename because ipfs.io DNS may change anytime.**
```
obiyankenobi, line 48: This is different than what we did with 9Block. We're using `https://ipfs.io/` there, so maybe it's a corner case we'll have to handle.

andreabadesso:
> The NFT creator will need to get in touch with Hathor team on Discord, in order to have the token identified as a NFT.

Maybe re-phrase it to: "in order to have the token identified as a NFT **on the official hathor explorer**"

text/0000-nft-standard.md
pedroferreira1: They will continue being identified as an NFT by us but the standard must be what we think is the best for new use cases, that's why we suggest having `ipfs://`

text/0000-nft-standard.md
```diff
@@ -0,0 +1,133 @@
...
+In Hathor Network an NFT is a new custom token created and its unique identifier is the transaction hash.
+
+## Transaction standard
+
+A transaction can be identified as an NFT creation if it has the following structure:
+
+1. Transaction version is 2 (TOKEN_CREATION_TRANSACTION).
+1. All inputs are HTR inputs.
+1. First output has HTR value and the script is a pushdata with the NFT data followed by a OP_CHECKSIG.
+1. We can have 1 or 0 melt authority output.
```
msbrogli, line 35: Suggestion: It can optionally have a melt authority output (but never more than one).

text/0000-nft-standard.md
```diff
@@ -0,0 +1,133 @@
...
+1. We can have 1 or 0 mint authority output.
```
msbrogli, line 36: Suggestion: It can optionally have a mint authority output (but never more than one).

text/0000-nft-standard.md
```diff
@@ -0,0 +1,133 @@
...
+1. We have N outputs with custom token created value (the value can be any number).
```
msbrogli, line 37: Suggestion: It has one or more outputs with the created token (any value is valid).

text/0000-nft-standard.md
```diff
@@ -0,0 +1,133 @@
...
+1. We can have 1 or 0 HTR output with change value.
```
msbrogli, line 38: Suggestion: It can optionally have an output for the HTR change.

text/0000-nft-standard.md
```diff
@@ -0,0 +1,133 @@
...
+
+Any transaction that has this structure will be identified as an NFT and the explorer/wallet screens will show specific NFT information for them.
```
msbrogli, line 40: Maybe we should say that the media associated with the NFT will not be show unless the NFT goes through Hathor Labs' review process.

text/0000-nft-standard.md
msbrogli: I agree with @pedroferreira1 . We manually marked 9Block's NFT as NFTs. But future NFTs should comply with this new standard.

text/0000-nft-standard.md
```diff
@@ -0,0 +1,133 @@
...
+        }
+    ]
+}
+```
+
+## Custom NFT
+
+There are some special cases where the NFT token won't follow the proposed standard, e.g. if it needs more than one data output. In that case, our wallets and explorer won't automatically identify this token as NFT.
+
+Given that this situation is expected to be rare, we will handle them manually. The NFT creator will need to get in touch with Hathor team on Discord, in order to have the token identified as an NFT on the official Hathor explorer. Besides that, if the digital asset URL follows the immutable requirements, it can be shown in the explorer just like any other standard NFT.
```
msbrogli, line 128: I'd remove the "on Discord" because it might change in the future.

msbrogli, line 128: Suggestion: Besides that, as long as the digital asset's URL is immutable, it should be approved in the review and should be shown in Hathor's Public Explorer like any other standard NFT.

text/0000-nft-standard.md
```diff
@@ -0,0 +1,133 @@
...
+
+# Prior art
+[prior-art]: #prior-art
+
+We were using the attributes on metadata JSON as an object but I changed to an array of objects after reading [OpenSea metadata standard](https://docs.opensea.io/docs/metadata-standards) to make it more flexible for future changes.
```
msbrogli, line 133: I don't think it is necessary. The goal here is to compare to other project and I think you can just cite OpenSea meta standard (or just say at the beginning of this document that it is based or inspired on the OpenSea metadata standard).

msbrogli (changes requested):

text/0000-nft-standard.md
pedroferreira1: Makes sense. Added it

text/0000-nft-standard.md
pedroferreira1: Yeah, makes sense. I removed the prior art section and mentioned the OpenSea standard in the metadata section

text/0000-nft-standard.md
```diff
@@ -0,0 +1,151 @@
...
+- 10 NFT units: deposit of 0.01 HTR.
+- 1 NFT unit: deposit of 0.01 HTR.
+
+This deposit may be returned in case of a melt of the created NFTs. Important to notice that you have the deposit back only if you melt an amount bigger than or equal to 100 units, just like the normal custom token.
+
+### Fee
+
+The first output of a NFT transaction will contain the script with the data string. This output must have some value and we use 0.01 HTR for that. It's an output that can never be returned.
+
+#### Why use HTR for the fee and not the custom token created?
```
msbrogli, line 143: I feel it should be moved to an FAQ section. It seems move like a curiosity than part of the standard.

msbrogli (approved):

obiyankenobi (approved):

PR #34, opened 2021-10-07
docs: private testnet guide v2

luislhl:
### Acceptance Criteria
- [ ] We should include how to run the Explorer
- [ ] We should use the new version of hathor-wallet-headless that allows configuring the tx-mining-url

Depends on the release of https://github.com/HathorNetwork/hathor-wallet-lib/pull/277 in the wallet-lib, and on releases of the Explorer and Wallet-headless to use the new wallet-lib.


luislhl:
@jansegre @msbrogli , could you review this?

jansegre (approved):
Looks great!

PR #37, opened 2021-11-11
docs: private testnet guide v2

luislhl:
### Acceptance Criteria
- [x] We should include how to run the Explorer
- [x] We should use the new version of hathor-wallet-headless that allows configuring the tx-mining-url
- [x] We should update the diagram, because we don't really use 2 wallets in the guide

Depends on the release of https://github.com/HathorNetwork/hathor-wallet-lib/pull/277 in the wallet-lib, and on releases of the Explorer and Wallet-headless to use the new wallet-lib.


jansegre (approved):

luislhl:
Merging this with approval from @msbrogli given in Slack

PR #40, opened 2021-12-07
Enhanced NFT metadata standard

Sharaddition:
Various additional properties are introduced in Metadata to improve the NFT standards on Hathor, such as Royalty & animation URL, without affecting the already minted NFTs.

### DIFF
--- a/text/0032-nft-standard.md
+++ b/text/0032-nft-standard.md
@@ -49,7 +49,7 @@ The most common use case is to represent a digital asset, then the NFT data usua
 
 ## Metadata
 
-Most NFTs also need extra data besides just the digital asset URL. The metadata standard is important to be followed in order to help future platform integrations with Hathor NFTs. This standard was inspired by [OpenSea metadata standard](https://docs.opensea.io/docs/metadata-standards).
+Most NFTs also need extra data besides just the digital asset URL. The metadata standard is important to be followed in order to help future platform integrations with Hathor NFTs. This standard was inspired by [OpenSea metadata standard](https://docs.opensea.io/docs/metadata-standards) and [Metaplex Token Metadata Standard](https://docs.metaplex.com/nft-standard).
 
 If the NFT requires a metadata, the output script data saved on blockchain **MUST** be the metadata URL in an immutable protocol (e.g. ipfs as explained in the last section), e.g. ipfs://ipfs/<hash>/metadata.json.
 
@@ -69,14 +69,43 @@ The metadata should have the following JSON structure:
         "type": "string",
         "description": "A URL pointing to an immutable resource with a digital asset to which this token represents"
     },
+    "collection": {
+        "type": Object{CollectionObject},
+        "description": "Additional details are recommended for NFT collections with multiple unique assets under one single family and symbol."
+    }
     "attributes": {
         "type": "Array<AttributeObject>",
         "description": "These are the extra attributes for the digital asset. It's an array of AttributeObject to make it as flexible as possible."
     },
+    "royalty": {
+        "type": "Object<RoyaltyObject>",
+        "description": "Required fields that would be utilised to pay royalty to original creators in secondary sales conducted by the exchange that adheres Hathor Metadata standards, after minting (primary sale)."
+    },
+    "external_info": {
+        "type": "Object{InfoObject}",
+        "description": "Additional information for collection, to be used by exchange or custom NFT assets, e.g. Gaming Assets"
+    }
 }
 ```
 
-The AttributeObject has the following JSON structure:
+- The CollectionObject has the following JSON structure:
+
+```
+"name": {
+    "type": "string",
+    "description": "Name of the NFT asset, e.g. Hippos #1"
+},
+"family": {
+    "type": "string",
+    "description": "Represents the family of the NFT collection, e.g. HathorLand DAO"
+},
+"symbol": {
+    "type": "string",
+    "description": "Symbol of the NFT asset within 2-5 characters, e.g. HIPPO"
+}
+```
+
+- The AttributeObject has the following JSON structure:
 
 ```
 {
@@ -93,6 +122,54 @@ The AttributeObject has the following JSON structure:
 
 The AttributeObject may have more attributes than the ones described above but those are the required ones.
 
+- The RoyaltyObject has the following JSON structure:
+
+```
+"creator_royalty_basis_points": {
+    "type": "integer",
+    "description": "Royalties percentage awarded to creators (basis points), e.g. 500"
+},
+"creators": {
+    "type": "Array<CreatorObject>",
+    "description": "Array of all creators involved in the project & their royalty bps."
+}
+```
+500 Basis Points mean 5% royalty, One basis point is equal to 1/100th of 1%. More info on BPS at [Investopedia](https://www.investopedia.com/terms/b/basispoint.asp)
+
+- The RoyaltyObject<**CreatorObject**> has the following JSON structure:
+
+```
+{
+    "address": {
+        "type": "string",
+        "description": "Address of the wallet in which royalty fees will be sent for secondary sales."
+    },
+    "share": {
+        "type": "integer",
+        "description": "Amount of royalty for each creator in bps, totaling a sum of 100."
+    }
+}
+```
+
+The CreatorObject may have more addresses than the ones described above, and the royalties will be split as per the share stated.
+
+- The InfoObject has the following JSON structure:
+
+```
+"animation_url": {
+    "type": "url",
+    "description": "URL to a non-immutable file hosted on CDN, given priority if the immutable file type is 3D object or isn't curently supported by the exchange or wallet preview."
+},
+"external_url": {
+    "type": "url",
+    "description": "URL to external website of the related project, e.g. Game website of the NFT asset."
+},
+"custom_fields": {
+    "type": "string | decimal",
+    "description": "URLs to social handles or preferably used for linking additional assets to a custom NFT token, e.g. Game Character Identifier, Discord Url"
+}
+```
+
 ### Metadata example
 
 ```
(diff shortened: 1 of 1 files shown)


text/0032-nft-standard.md
```diff
@@ -69,14 +69,43 @@ The metadata should have the following JSON structure:
...
+        "type": "Object{InfoObject}",
+        "description": "Additional information for collection, to be used by exchange or custom NFT assets, e.g. Gaming Assets"
+    }
+}
+```
+
+- The CollectionObject has the following JSON structure:
+
+```
+"name": {
```
obiyankenobi, line 94: Why are we repeating the name here? Is it standard in some other NFT blockchains?

Sharaddition: Yeah, Solana metadata standards are also repeating name prop. So thought to follow the same.

jansegre:
Currently NFTs metadata don't require any intepretation by the full-node (hathor-core). A royalty mechanism definitely would require interpretation in order to be enforced (otherwise if only the wallet supports it, it's just a suggestion). This type of change on the full-node would need at least soft-fork and a migration path, so it isn't by any means simple to achieve.

I definitely think such mechanism has its usefulness, but I don't believe this changes cover what we actually need to support royalties. I'd at least expect it to be its own RFC. But anyway I don't believe this is the best medium to propose this change, I'd expect a lot of back and forth _before_ we start an RFC, and it probably should be it's own RFC.

About the Metaplex token standard, that looks much easier to support, unless it requires specific mechanisms like royalties.

Sharaddition:
> A royalty mechanism definitely would require interpretation in order to be enforced (otherwise if only the wallet supports it, it's just a suggestion).

Exchanges that need to pay royalty can conduct an off-chain split of the amount, till Nano Contracts arrive.

obiyankenobi:
> Currently NFTs metadata don't require any intepretation by the full-node (hathor-core). A royalty mechanism definitely would require interpretation in order to be enforced (otherwise if only the wallet supports it, it's just a suggestion). This type of change on the full-node would need at least soft-fork and a migration path, so it isn't by any means simple to achieve.

@jansegre I don't think it was a suggestion that this was enforced on the full node. But having on the metadata makes it easier for different marketplaces to enforce the same royalty. I believe this is similar to what happens in other chains.

jansegre:
I understand marketplaces are much more likely to honor the metadata. But if it's acceptable for OTC sales to skip royalties, then it's OK.

text/0032-nft-standard.md
```diff
@@ -69,14 +69,43 @@ The metadata should have the following JSON structure:
...
         "type": "Array<AttributeObject>",
         "description": "These are the extra attributes for the digital asset. It's an array of AttributeObject to make it as flexible as possible."
     },
+    "royalty": {
+        "type": "Object<RoyaltyObject>",
+        "description": "Required fields that would be utilised to pay royalty to original creators in secondary sales conducted by the exchange that adheres Hathor Metadata standards, after minting (primary sale)."
+    },
+    "external_info": {
+        "type": "Object{InfoObject}",
+        "description": "Additional information for collection, to be used by exchange or custom NFT assets, e.g. Gaming Assets"
```
pedroferreira1, line 86: The additional information is not only for the collection, it may be from the NFT, or the owner website. Maybe we can have "Additional information with external links for the NFT or collection."

text/0032-nft-standard.md
```diff
@@ -69,14 +69,43 @@ The metadata should have the following JSON structure:
...
+    }
+}
+```
+
+- The CollectionObject has the following JSON structure:
+
+```
```
pedroferreira1, line 93: To keep the same standard as in the `AttributeObject`, we must have `{` to start the object and `}` to close it. Similar to what you did in the `CreatorObject`.

text/0032-nft-standard.md
```diff
@@ -93,13 +122,66 @@ The AttributeObject has the following JSON structure:
 
 The AttributeObject may have more attributes than the ones described above but those are the required ones.
 
+- The RoyaltyObject has the following JSON structure:
+
+```
```
pedroferreira1, line 127: To keep the same standard as in the `AttributeObject`, we must have `{` to start the object and `}` to close it. Similar to what you did in the `CreatorObject`.

text/0032-nft-standard.md
```diff
@@ -69,14 +69,43 @@ The metadata should have the following JSON structure:
...
+    "type": "string",
+    "description": "Name of the NFT asset, e.g. Hippos #1"
+},
+"family": {
+    "type": "string",
+    "description": "Represents the family of the NFT collection, e.g. HathorLand DAO"
+},
+"symbol": {
+    "type": "string",
+    "description": "Symbol of the NFT asset within 2-5 characters, e.g. HIPPO"
```
pedroferreira1, line 104: I think this is the collection symbol, not the NFT symbol

text/0032-nft-standard.md
```diff
@@ -93,13 +122,66 @@ The AttributeObject has the following JSON structure:
 
 The AttributeObject may have more attributes than the ones described above but those are the required ones.
 
+- The RoyaltyObject has the following JSON structure:
+
+```
+"creator_royalty_basis_points": {
+    "type": "integer",
+    "description": "Royalties percentage awarded to creators (basis points), e.g. 500"
```
pedroferreira1, line 130: I think would be good to add the % here as well. "..., e.g. 500, which means 5%"

text/0032-nft-standard.md
```diff
@@ -93,13 +122,66 @@ The AttributeObject has the following JSON structure:
...
+```
+
+The CreatorObject may have more addresses than the ones described above, and the royalties will be split as per the share stated.
+
+- The InfoObject has the following JSON structure:
+
+```
+"animation_url": {
+    "type": "url",
+    "description": "URL to a non-immutable file hosted on CDN, given priority if the immutable file type is 3D object or isn't curently supported by the exchange or wallet preview."
```
pedroferreira1, line 161: typo currently*

text/0032-nft-standard.md
```diff
@@ -93,13 +122,66 @@ The AttributeObject has the following JSON structure:
...
+},
+"external_url": {
+    "type": "url",
+    "description": "URL to external website of the related project, e.g. Game website of the NFT asset."
+},
+"custom_fields": {
```
pedroferreira1, line 167: Why is this relevant? Anyone can add anything to the JSON metadata, the standard is just to create a model so anyone integrating with the NFTs can know how to find some information. If we feel those custom fields are important for the marketplaces, then I feel we must have an object with them, not a free field without a clear description. 

In the example below you don't have the key `custom_fields` but you have

```
       "dna": "6c191e41-5bfe-432f-9e4b-ddcc718dce92",
       "discord": "https://discord.gg/87bYu52gDW"
```

Maybe they could be

```
"custom_information": {
       "dna": "6c191e41-5bfe-432f-9e4b-ddcc718dce92",
       "discord": "https://discord.gg/87bYu52gDW"
}
```

So it's easier for integrations.

text/0032-nft-standard.md
```diff
@@ -93,13 +122,66 @@ The AttributeObject has the following JSON structure:
...
+
 ### Metadata example
 
 ```
 {
     "name": "Gandalf",
     "description": "A wizard, one of the Istari order, and the leader and mentor of the Fellowship of the Ring",
     "file": "ipfs://ipfs/QmbuthvFV2EjvfmWXxt2L83PwPPwbjjggBhVsrEB7AXW123/gandalf.png",
+    "collection": {
+       "name": "Gandalf",
```
pedroferreira1, line 181: If we change the description, then we should change it here as well.

text/0032-nft-standard.md
```diff
@@ -117,7 +199,30 @@ The AttributeObject may have more attributes than the ones described above but t
             "type": "ring",
             "value": 0
         }
-    ]
+    ],
+   "royalty": {
```
pedroferreira1, line 203: The indentation of this object should be fixed.

text/0032-nft-standard.md
obiyankenobi: It's not clear to me why `external_info` is needed. I don't get the point of `animation_url` and `external_url`, and `custom_fields` seems similar to what we already have in `attributes`.

text/0032-nft-standard.md
```diff
@@ -117,7 +199,30 @@ The AttributeObject may have more attributes than the ones described above but t
...
+                "address": "H7t2eNhFNeH3hrUAJj4AAu9amtWmPT4fLB",
+                "share": 95
+              },
+              {
+                "address": "HNBuYEsJYznh5zZGx4eDn2NXYuiPpgKBgH",
+                "share": 5
+              },
+              {
+                "address": "RaNd0MsJYznh5zYOyoDn2NXYuiPAdDre5s",
+                "share": 0
```
obiyankenobi, line 216: In this case, the idea is just to identify this address as a creator, even if it's not receiving any royalties?

text/0032-nft-standard.md
```diff
@@ -146,8 +251,16 @@ There are some special cases where the NFT token won't follow the proposed stand
 
 Given that this situation is expected to be rare, we will handle them manually. The NFT creator will need to get in touch with Hathor team, in order to have the token identified as an NFT on the official Hathor explorer. Besides that, as long as the digital asset's URL is immutable, it should be approved in the review and should be shown in Hathor's Public Explorer like any other standard NFT.
 
+## Drawbacks
+
+- The main drawback is increased storage usage. File to be stored on immutable storage will cost more than prior standard.
```
obiyankenobi, line 256: These should be drawbacks of the metadata standard, not this new version.

text/0032-nft-standard.md
```diff
@@ -146,8 +251,16 @@ There are some special cases where the NFT token won't follow the proposed stand
 
 Given that this situation is expected to be rare, we will handle them manually. The NFT creator will need to get in touch with Hathor team, in order to have the token identified as an NFT on the official Hathor explorer. Besides that, as long as the digital asset's URL is immutable, it should be approved in the review and should be shown in Hathor's Public Explorer like any other standard NFT.
 
+## Drawbacks
```
obiyankenobi, line 254: I would remove this section entirely. Drawbacks for this version should be in the comments of this PR, not the final RFC.

text/0032-nft-standard.md
```diff
@@ -69,14 +69,43 @@ The metadata should have the following JSON structure:
         "type": "string",
         "description": "A URL pointing to an immutable resource with a digital asset to which this token represents"
     },
+    "collection": {
```
obiyankenobi, line 72: Should we mark the fields that are required or optional? This would be good now that the standard has gotten more complex.

pedroferreira1:
Done in https://github.com/HathorNetwork/rfcs/pull/43

PR #44, opened 2022-09-22
feat: add security best practices for use cases RFC

pedroferreira1:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/8aac52dc23ffffe9251b1313c1117ec1343d3e42/text/0000-use-case-integration-best-practices.md)

### DIFF
--- a/text/0000-use-case-integration-best-practices.md
+++ b/text/0000-use-case-integration-best-practices.md
@@ -46,11 +46,11 @@ There are multiple ways that a use case can integrate with Hathor, here is a com
 
 We strongly recommend use cases to run two or more full nodes as a protection to direct attacks to their full nodes.
 
-Use case's full nodes **should not** be connected among them. This is important to mitigate some attack vectors. Remember that the transactions will be propagate by the p2p network and all use case's full nodes will receive the transactions eventually during normal network activity.
+Use case's full nodes **should not** be connected among them. This is important to mitigate some attack vectors. Remember that the transactions will be propagated by the p2p network and all use case's full nodes will receive the transactions eventually during normal network activity.
 
 ### Validate new transactions on more than one full node before accepting them
 
-Let's assume an exchange wants to run nodes to identify deposits in the Hathor network, so a recommended approach for the integration would be to run at least two full nodes (node1 and node2), which are not connected between them (node1 must have node2 in the blacklist and vice versa). In that architecture, if any deposit is identified in node1, then the exchange must check that it's also a valid transaction in node2 and in one of the public nodes. With this approach, if one of the nodes is compromised and with a different consensus/state, an alert must be raised, so we must investigate what happened.
+Let's assume an exchange wants to run nodes to identify deposits in the Hathor network, so a recommended approach for the integration would be to run at least two full nodes (node1 and node2), which are not connected between them (node1 must have node2 in the blacklist and vice versa). In that architecture, if any deposit is identified in node1, then the exchange must check that it's also a valid transaction in node2 and in one of the public nodes. In this approach, if an attacker successfully compromises one of your full nodes, your validation would fail and the deposit will not be accepted.
 
 ### Validate all your full nodes have the same best block
 
@@ -59,11 +59,9 @@ Use cases should regularly check whether the best block is the same on all their
 
 ## Peer-id
 
-The peer-id is a unique identifier of your full node in Hathor's p2p network. You must keep your peer-id secret to prevent attackers from directly targeting your full nodes. Do not tell anyone your peer-ids, and do not publish them on public channels, only in private groups with the Hathor team. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.
+The peer-id is a unique identifier of your full node in Hathor's p2p network. You must keep your peer-id secret to prevent attackers from directly targeting your full nodes. Do not tell anyone your peer-ids, and do not publish them on public channels. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.
 
-## Be alert for weird behavior
-
-### How to validate a new transaction
+## How to validate a new transaction
 
 The transactions in the Hathor network have many fields that must be checked to guarantee that a transaction is valid for your use case. For more details about the fields of a transaction, check the [Transaction Anatomy RFC](https://github.com/HathorNetwork/rfcs/blob/master/text/0015-anatomy-of-tx.md).
 
@@ -75,37 +73,35 @@ The transactions in the Hathor network have many fields that must be checked to 
   - [Address](#address)
   - [Timelock](#timelock)
 
-#### Version
+### Version
 
-Transactions have `version=1`, blocks have versions `0` or `3`, and token creation transactions have `version=2`.
+Version identifies the type of vertex. The possible choices are:
 
-#### Voided state
+- Block: 0
+- Regular Transaction: 1
+- Token Creation Transaction: 2
+- Merged Mined Block: 3
 
-A voided transaction is **not** a valid transaction. You must only accept transactions that have `is_voided=false`.
+Depending on your use case you must accept one or more types.
 
-#### Outputs
+### Voided state
 
-The outputs contain the fields used to identify that the funds belong to your wallet.
+A voided transaction is cancelled and should **never** be accepted. You must validate that the transaction is not voided asserting that `is_voided == false`.
 
-##### Token
+### Outputs
 
-The output has the token information, so you can check if it has the expected tokens. For example, exchanges expect the deposits to be of HTR, so `token = '00'` and `token_data=0`.
+Hathor supports multi-token transactions. You must confirm that your outputs are correctly placed as follows:
 
-##### Value
+- Token id is correct. If you accept only HTR token, token id must be `"00"`.
+- Token data is equal to `0`, if it's HTR token.
+- Value matches the expected value. Note that it is an integer and `12.34` is represented by `1234`.
+- Timelock must be `null`; otherwise your funds might be locked.
 
-The value of the output is an integer, so `1000` means `10.00`.
+### Number of confirmations
 
-##### Address
+Some use cases might handle transactions with huge amounts, so it's essential to wait for some blocks to confirm the transaction before accepting it as a valid one. The more blocks confirm a transaction, the more guarantee there is that this transaction won't become voided in the future. As a reference, Bitcoin's use cases usually require six confirmations before accepting a new deposit.
 
-The output address must be one of your wallet's addresses.
-
-##### Timelock
-
-The output must have a timelock, and if it's bigger than the current timestamp, you won't be able to spend this output until this moment is reached. Because of that, it's essential to validate that `timelock=null` before validating this transaction.
-
-#### Number of confirmations
-
-Some use cases might handle transactions with huge amounts, so it's essential to wait for some blocks to confirm the transaction before accepting it as a valid one. The more blocks confirm a transaction, the more guarantee there is that this transaction won't become invalid in the future. As a reference, Bitcoin's use cases usually require a six confirmations before accepting a new deposit.
+## Be alert for weird behavior
 
 ### Check if an unusual amount of deposits or withdrawals are being made
 
@@ -113,11 +109,11 @@ Many use cases have withdrawals/deposits in their set of features through blockc
 
 For that situation, it's crucial to have an easy way to block specific accounts that have unusual behavior and might be part of an attack. You might also consider to limit the number of operations a user can do in a time window.
 
-This validation and user block must be done in the use-case software application.
+This validation and user throttling must be done in the use-case software application.
 
 ### Check if one of your full nodes gets out-of-sync
 
-One other relevant aspect of full node is to always check for weird behavior is the sync among them. We recommend use cases to regularly validate that all their full nodes are in sync among them and in sync with at least one public node as well.
+Always check for weird behavior in the synchronization among full nodes. We recommend use cases to regularly validate that all their full nodes are in sync among them and in sync with at least one public node as well.
 
 This validation is important to guarantee the node is not isolated from the rest of the network with a fork of the blockchain.
 
--- a/text/0000-use-case-integration-best-practices.md
+++ b/text/0000-use-case-integration-best-practices.md
@@ -8,43 +8,16 @@
 # Summary
 [summary]: #summary
 
-This document presents best practices for use cases that integrate with the Hathor Network, in order to improve the reliability of their operation.
+This document presents security best practices for use cases that integrate with the Hathor Network, in order to improve the reliability of their operation.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-There are multiple ways that a use case can integrate with Hathor, here is a basic architecture of a common integration.
+There are multiple ways that a use case can integrate with Hathor, here is a common architecture compliant to our recommendations
 
 - The use case has a software that runs operations offchain and is used by its customer.
-- This software connects with a headless wallet, that executes operations in the Hathor network.
-- The headless wallet is connected to a full node, that belongs to the Hathor peer-to-peer network.
-
-<pre>
-┌─────────────┐      ┌──────────┐                     ┌─────────────┐        ┌─────────────┐
-│             │      │          │◀────────────────────│             │        │             │
-│  Hathor     │      │ Headless │                     │  Use Case   │        │  Customer   │
-│  Full Node  │◀────▶│  Wallet  │                     │  Software   │◀──────▶│             │
-│             │      │          │────────────────────▶│             │        │             │
-└─────────────┘      └──────────┘                     └─────────────┘        └─────────────┘
-</pre>
-
-## Run more than one node
-
-Even though the architecture described above works fine for a simple use case, we strongly recommend any use case to run more than one node, in order to increase the reliability of the operation.
-
-During normal network activity, any transaction will be propagated through all peers and they will get to the same consensus. However, there are some situations where this might not be true, e.g. if an attack is happening to a single peer. To improve the reliability of the operation at any moment, it's important that all use cases run more than one full node and they **can't be connected among them**, only to the rest of the network.
-
-Let's assume an exchange wants to run nodes to identify deposits in the Hathor network, so a recommended approach for the integration would be to run at least two full nodes (node1 and node2), which are not connected between them (node1 must have node2 in the blacklist and vice versa). In that architecture, if any deposit is identified in node1, then the exchange must check that it's also a valid transaction in node2 and in one of the public nodes. With this approach, if one of the nodes is compromised and with a different consensus/state, an alert must be raised, so we must investigate what happened.
-
-### Validate new transactions on more than one full node before accepting them
-
-When running more than one node, any new transaction must be accepted only after validating that it's a valid transaction in all the nodes.
-
-### Validate all your full nodes have the same best block
-
-It's also important to regularly check if all nodes have the same best block. If they have different best blocks, this might be a clue that one of the nodes could be a target of an attack because is running a separate fork of the blockchain.
-
-### Recommended architecture
+- This software connects to a headless wallet which communicates with a full node. The full node is connected to the Hathor's p2p network.
+- The headless wallet will handle all events from the full node to keep its balance and transaction history updated. The headless software have APIs to manage a wallet, e.g. get addresses, create and push transactions to the network, create new tokens, and some other features described in the project [repository](https://github.com/HathorNetwork/hathor-wallet-headless).
 
 <pre>
                        ┌─────────────┐      ┌──────────┐                     ┌─────────────┐        ┌─────────────┐
@@ -69,23 +42,44 @@ It's also important to regularly check if all nodes have the same best block. If
 
 </pre>
 
+## Run more than one node
+
+We strongly recommend use cases to run two or more full nodes as a protection to direct attacks to their full nodes.
+
+Use case's full nodes **should not** be connected among them. This is important to mitigate some attack vectors. Remember that the transactions will be propagate by the p2p network and all use case's full nodes will receive the transactions eventually during normal network activity.
+
+### Validate new transactions on more than one full node before accepting them
+
+Let's assume an exchange wants to run nodes to identify deposits in the Hathor network, so a recommended approach for the integration would be to run at least two full nodes (node1 and node2), which are not connected between them (node1 must have node2 in the blacklist and vice versa). In that architecture, if any deposit is identified in node1, then the exchange must check that it's also a valid transaction in node2 and in one of the public nodes. With this approach, if one of the nodes is compromised and with a different consensus/state, an alert must be raised, so we must investigate what happened.
+
+### Validate all your full nodes have the same best block
+
+Use cases should regularly check whether the best block is the same on all their full nodes. If full node have different best blocks, the validation must be done again some seconds later because this might happen depending on the network block propagation time. If the differente continue, the nodes might be under attack and the use case should consider blocking deposits.
+
+
 ## Peer-id
 
-The peer-id is a unique identifier of your full node in Hathor's p2p network. Even though it does not contain any private key information, keeping it secret is a security measure that prevents attackers from targeting directly your full nodes.
-
-Because of that, it's important to never share it in public chats, only in private groups with the Hathor team. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.
+The peer-id is a unique identifier of your full node in Hathor's p2p network. You must keep your peer-id secret to prevent attackers from directly targeting your full nodes. Do not tell anyone your peer-ids, and do not publish them on public channels, only in private groups with the Hathor team. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.
 
 ## Be alert for weird behavior
 
-### Correctly validate transactions
+### How to validate a new transaction
 
 The transactions in the Hathor network have many fields that must be checked to guarantee that a transaction is valid for your use case. For more details about the fields of a transaction, check the [Transaction Anatomy RFC](https://github.com/HathorNetwork/rfcs/blob/master/text/0015-anatomy-of-tx.md).
+
+- [Version](#version)
+- [Voided state](#voided-state)
+- Output
+  - [Token](#token)
+  - [Value](#value)
+  - [Address](#address)
+  - [Timelock](#timelock)
 
 #### Version
 
 Transactions have `version=1`, blocks have versions `0` or `3`, and token creation transactions have `version=2`.
 
-#### Voided
+#### Voided state
 
 A voided transaction is **not** a valid transaction. You must only accept transactions that have `is_voided=false`.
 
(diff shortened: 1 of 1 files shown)


text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+The output address must be one of your wallet's addresses.
+
+##### Timelock
+
+The output must have a timelock, and if it's bigger than the current timestamp, you won't be able to spend this output until this moment is reached. Because of that, it's essential to validate that `timelock=null` before validating this transaction.
+
+#### Number of confirmations
+
+Some use cases might handle transactions with huge amounts, so it's essential to wait for some blocks to confirm the transaction before accepting it as a valid one. The more blocks confirm a transaction, the more guarantee there is that this transaction won't become invalid in the future.
```
luislhl, line 114: Does it make sense to suggest how many blocks it makes sense to wait for a level of security similar to waiting 6 blocks in Bitcoin (or some other comparative measure)?

luislhl, line 114: Should we detail in the reference-level how to check if a transaction has already been confirmed by more than N blocks?

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+- Status: Draft
+- Start Date: 2022-09-20
+- RFC PR: (leave this empty)
+- Hathor Issue: (leave this empty)
+- Author: Pedro Ferreira <pedro@hathor.network>
+
+# Summary
+[summary]: #summary
+
+This document presents best practices for use cases that integrate with the Hathor Network, in order to improve the reliability of their operation.
```
luislhl, line 11: The goal here is to present general best-practices or security best-practices?

In the first case, I think we might have some other best-practices that could be linked here, like this one which is still unmerged: https://github.com/HathorNetwork/rfcs/pull/30, or even something about the tx-mining-service.

But if we are focused on security here, maybe we should explicitly state this in the title and summary.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+The output must have a timelock, and if it's bigger than the current timestamp, you won't be able to spend this output until this moment is reached. Because of that, it's essential to validate that `timelock=null` before validating this transaction.
+
+#### Number of confirmations
+
+Some use cases might handle transactions with huge amounts, so it's essential to wait for some blocks to confirm the transaction before accepting it as a valid one. The more blocks confirm a transaction, the more guarantee there is that this transaction won't become invalid in the future.
+
+### Check if an unusual amount of deposits or withdrawals are being made
+
+Many use cases have withdrawals/deposits in their set of features through blockchain transactions. It's important to be alerted in case any unusual amount of deposits/withdrawals are being made because this could be caused by an attack.
```
luislhl, line 118: This would be done at the "use-case software" level, right? Not by checking something in the fullnode or in the wallet-headless.

If so, maybe we should make this clear.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+- The use case has a software that runs operations offchain and is used by its customer.
+- This software connects with a headless wallet, that executes operations in the Hathor network.
+- The headless wallet is connected to a full node, that belongs to the Hathor peer-to-peer network.
+
+<pre>
+┌─────────────┐      ┌──────────┐                     ┌─────────────┐        ┌─────────────┐
+│             │      │          │◀────────────────────│             │        │             │
+│  Hathor     │      │ Headless │                     │  Use Case   │        │  Customer   │
+│  Full Node  │◀────▶│  Wallet  │                     │  Software   │◀──────▶│             │
```
msbrogli, line 26: If we will recommend one to use more than one full node, I feel we should add the second full node in this diagram. We will explain the reasons behind it later.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+# Summary
+[summary]: #summary
+
+This document presents best practices for use cases that integrate with the Hathor Network, in order to improve the reliability of their operation.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+There are multiple ways that a use case can integrate with Hathor, here is a basic architecture of a common integration.
```
msbrogli, line 16: We can say that `here is a common architecture compliant to our recommendations`.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+│             │      │          │◀────────────────────│             │        │             │
+│  Hathor     │      │ Headless │                     │  Use Case   │        │  Customer   │
+│  Full Node  │◀────▶│  Wallet  │                     │  Software   │◀──────▶│             │
+│             │      │          │────────────────────▶│             │        │             │
+└─────────────┘      └──────────┘                     └─────────────┘        └─────────────┘
+</pre>
+
+## Run more than one node
+
+Even though the architecture described above works fine for a simple use case, we strongly recommend any use case to run more than one node, in order to increase the reliability of the operation.
```
msbrogli, line 33: I feel we should just say "we strongly recommend use cases to run two or more full nodes as a protection to direct attacks to their full nodes".

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+During normal network activity, any transaction will be propagated through all peers and they will get to the same consensus. However, there are some situations where this might not be true, e.g. if an attack is happening to a single peer. To improve the reliability of the operation at any moment, it's important that all use cases run more than one full node and they **can't be connected among them**, only to the rest of the network.
```
msbrogli, line 35: I feel it is confusing. A suggestion would be to start with the recommendation itself and explain it later. For example:

Use case's full nodes **should not** be connected among them. This is important to mitigate some attack vectors. Remember that the transactions will be propagate by the p2p network and all use case's full nodes will receive the transactions eventually.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+Let's assume an exchange wants to run nodes to identify deposits in the Hathor network, so a recommended approach for the integration would be to run at least two full nodes (node1 and node2), which are not connected between them (node1 must have node2 in the blacklist and vice versa). In that architecture, if any deposit is identified in node1, then the exchange must check that it's also a valid transaction in node2 and in one of the public nodes. With this approach, if one of the nodes is compromised and with a different consensus/state, an alert must be raised, so we must investigate what happened.
```
msbrogli, line 37: Should we add it to the wallet headless? Anyway, I would move it to the next recommendation (validate new transactions on more than one full node).

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Validate new transactions on more than one full node before accepting them
+
+When running more than one node, any new transaction must be accepted only after validating that it's a valid transaction in all the nodes.
```
msbrogli, line 41: I feel we should explain it a bit more. For example: "When a new deposit is detected, this deposit should be accepted only after checking the following on all full nodes:

a) The transaction exists.
b) The transaction is not voided.
c) The transaction's first block is the same for all full nodes.
d) The full node's best block is the same for all full nodes.

If any condition fails, you should check again later."

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Validate all your full nodes have the same best block
+
+It's also important to regularly check if all nodes have the same best block. If they have different best blocks, this might be a clue that one of the nodes could be a target of an attack because is running a separate fork of the blockchain.
```
msbrogli, line 45: Suggestion: "Use cases should regularly check whether the best block is the same on all their full nodes. If full node have different best blocks, they might be under attack and should consider blocking deposits."

I feel we should improve this recommendation because a naive implementation that collects the best block from all full nodes and compare might fail due to a race condition. For example, a new block found during the data collection.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Recommended architecture
+
+<pre>
+                       ┌─────────────┐      ┌──────────┐                     ┌─────────────┐        ┌─────────────┐
```
msbrogli, line 50: I feel this should be the diagram at the beginning of the document. It seems confusing to have two different diagrams unless we have a section of "what not to do".

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+                       │  Hathor     │
+                       │ Full Node 2 │
+                       │             │
+                       └─────────────┘
+
+</pre>
+
+## Peer-id
+
+The peer-id is a unique identifier of your full node in Hathor's p2p network. Even though it does not contain any private key information, keeping it secret is a security measure that prevents attackers from targeting directly your full nodes.
```
msbrogli, line 74: I feel we should very clearly state the recommendation: "You must keep your peer-id secret to prevent attackers from targeting directly your full nodes. Do not tell anyone your peer-ids and do not publish them on public channels.".

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+Because of that, it's important to never share it in public chats, only in private groups with the Hathor team. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.
```
msbrogli, line 76: I'd remove the part about public chats, but I'd keep the "If you think your peer-id has been exposed..." part.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Be alert for weird behavior
```
msbrogli, line 78: I'd rename it to "How to validate a new transaction".

text/0000-use-case-integration-best-practices.md
msbrogli: I just found the "how to validate a new transaction" section. Maybe this should be moved there.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Correctly validate transactions
+
+The transactions in the Hathor network have many fields that must be checked to guarantee that a transaction is valid for your use case. For more details about the fields of a transaction, check the [Transaction Anatomy RFC](https://github.com/HathorNetwork/rfcs/blob/master/text/0015-anatomy-of-tx.md).
+
+#### Version
+
+Transactions have `version=1`, blocks have versions `0` or `3`, and token creation transactions have `version=2`.
```
msbrogli, line 86: I feel we should explain it better. Should use cases verify version? How?

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+The peer-id is a unique identifier of your full node in Hathor's p2p network. Even though it does not contain any private key information, keeping it secret is a security measure that prevents attackers from targeting directly your full nodes.
+
+Because of that, it's important to never share it in public chats, only in private groups with the Hathor team. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.
+
+## Be alert for weird behavior
+
+### Correctly validate transactions
+
+The transactions in the Hathor network have many fields that must be checked to guarantee that a transaction is valid for your use case. For more details about the fields of a transaction, check the [Transaction Anatomy RFC](https://github.com/HathorNetwork/rfcs/blob/master/text/0015-anatomy-of-tx.md).
```
msbrogli, line 82: I guess we should add a checklist here with all things that must be verified. Then, we can add markers [1], [2], [3], and add notes below this checklist.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### Version
+
+Transactions have `version=1`, blocks have versions `0` or `3`, and token creation transactions have `version=2`.
+
+#### Voided
```
msbrogli, line 88: I feel we are just explaining the meaning of each field. We can do it but we should also give our recommendation on how to validate each field.

text/0000-use-case-integration-best-practices.md
msbrogli: We might say "As a reference, Bitcoin's use cases usually require a six confirmations before accepting a new deposit."

text/0000-use-case-integration-best-practices.md
msbrogli, line 118: I'd say "check for unusual levels of deposits and withdrawals". You might also consider to limit the number of operations a user can do in a time window.

msbrogli (changes requested):

text/0000-use-case-integration-best-practices.md
pedroferreira1: Added this sentence

text/0000-use-case-integration-best-practices.md
pedroferreira1: Campana is now working on an integration guide for use cases. I feel we should focus this RFC in the security best practices and link them, as @msbrogli said. I will make this clear in the title and summary.

text/0000-use-case-integration-best-practices.md
pedroferreira1: Makes sense! Added a sentence to make this clear.

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+There are multiple ways that a use case can integrate with Hathor, here is a basic architecture of a common integration.
+
+- The use case has a software that runs operations offchain and is used by its customer.
+- This software connects with a headless wallet, that executes operations in the Hathor network.
+- The headless wallet is connected to a full node, that belongs to the Hathor peer-to-peer network.
+
```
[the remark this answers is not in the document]
pedroferreira1: Added

text/0000-use-case-integration-best-practices.md
pedroferreira1: Moved the second architecture here, as suggested in another comment

text/0000-use-case-integration-best-practices.md
pedroferreira1: Moved it to the beginning of the document

text/0000-use-case-integration-best-practices.md
pedroferreira1: Changed it, especially now that we have a single architecture scheme

text/0000-use-case-integration-best-practices.md
pedroferreira1: Yeah, it's better, changed it.

text/0000-use-case-integration-best-practices.md
pedroferreira1: You mean to add a new API in the headless to check a transaction in different nodes?

text/0000-use-case-integration-best-practices.md
pedroferreira1: Removed this sentence from this place.

text/0000-use-case-integration-best-practices.md
pedroferreira1: Improved the sentence to suggest executing more validations because of the block propagation time difference.

text/0000-use-case-integration-best-practices.md
pedroferreira1: Added

text/0000-use-case-integration-best-practices.md
pedroferreira1: I kept the private groups part, otherwise we would just be saying never to share it, however we need it for the whitelist. It's like this the section now:

> The peer-id is a unique identifier of your full node in Hathor's p2p network. You must keep your peer-id secret to prevent attackers from directly targeting your full nodes. Do not tell anyone your peer-ids, and do not publish them on public channels, only in private groups with the Hathor team. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.

What do you think?

text/0000-use-case-integration-best-practices.md
pedroferreira1: The "Be alert for weird behavior" has 3 sub sections:

- Correctly validate transactions
- Check if an unusual amount of deposits or withdrawals are being made
- Check if one of your full nodes gets out-of-sync

I changed the first one to "How to validate a new transaction" as you suggested

text/0000-use-case-integration-best-practices.md
pedroferreira1: That's one thing that I didn't like in this RFC, it doesn't describe in details the integration, so the answer for your "How" is: depends. 

If they are using the headless, is one way, if they are using the full node directly, is another way.

I could mention the exchange integration RFC (https://github.com/HathorNetwork/rfcs/blob/master/text/0027-exchange-integration.md#how-to-check-if-a-transaction-is-a-valid-deposit) and remove the whole "How to validate a transaction" section here, there we have all the API details. 

What do you think?

text/0000-use-case-integration-best-practices.md
pedroferreira1: The text is:

>  A voided transaction is **not** a valid transaction. You must only accept transactions that have `is_voided=false`.

I didn't understand your comment about the voided section

text/0000-use-case-integration-best-practices.md
pedroferreira1: Changed

text/0000-use-case-integration-best-practices.md
pedroferreira1: Added

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,177 @@
...
+### Check if an unusual amount of deposits or withdrawals are being made
+
+Many use cases have withdrawals/deposits in their set of features through blockchain transactions. It's important to be alerted in case any unusual amount of deposits/withdrawals are being made because this could be caused by an attack.
+
+For that situation, it's crucial to have an easy way to block specific accounts that have unusual behavior and might be part of an attack.
+
+### Check if one of your full nodes gets out-of-sync
+
+One other relevant aspect of full node to always check for weird behavior is the sync among them. We recommend use cases to regularly validate that all their full nodes are in sync among them and in sync with at least one public node as well.
```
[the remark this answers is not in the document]
pedroferreira1: Yeah, I agree it's important. Added a sentence about that as well.

text/0000-use-case-integration-best-practices.md
pedroferreira1: There's another comment from @msbrogli that I mention the exchange integration guide. (https://github.com/HathorNetwork/rfcs/blob/master/text/0027-exchange-integration.md#how-to-check-if-a-transaction-is-a-valid-deposit). There we also have some more details about validating the number of confirmation blocks.

I tried not to go deep into implementation details, because it will depend on the use case. Maybe I could just reference this guide in this RFC, what do you think?

luislhl: Sure, I think referencing this is enough

text/0000-use-case-integration-best-practices.md
```diff
@@ -69,23 +42,44 @@ It's also important to regularly check if all nodes have the same best block. If
 
 </pre>
 
-## Peer-id
+## Run more than one node
+
+We strongly recommend use cases to run two or more full nodes as a protection to direct attacks to their full nodes.
+
+Use case's full nodes **should not** be connected among them. This is important to mitigate some attack vectors. Remember that the transactions will be propagate by the p2p network and all use case's full nodes will receive the transactions eventually during normal network activity.
```
luislhl, line 49: propagate -> propagated

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,175 @@
...
+
+The output address must be one of your wallet's addresses.
+
+##### Timelock
+
+The output must have a timelock, and if it's bigger than the current timestamp, you won't be able to spend this output until this moment is reached. Because of that, it's essential to validate that `timelock=null` before validating this transaction.
+
+#### Number of confirmations
+
+Some use cases might handle transactions with huge amounts, so it's essential to wait for some blocks to confirm the transaction before accepting it as a valid one. The more blocks confirm a transaction, the more guarantee there is that this transaction won't become invalid in the future. As a reference, Bitcoin's use cases usually require a six confirmations before accepting a new deposit.
```
luislhl, line 108: "As a reference, Bitcoin's use cases usually require a six confirmations before accepting a new deposit."

Shouldn't we drop the "a" from "a six confirmations"?

luislhl (approved):

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Run more than one node
+
+We strongly recommend use cases to run two or more full nodes as a protection to direct attacks to their full nodes.
+
+Use case's full nodes **should not** be connected among them. This is important to mitigate some attack vectors. Remember that the transactions will be propagate by the p2p network and all use case's full nodes will receive the transactions eventually during normal network activity.
+
+### Validate new transactions on more than one full node before accepting them
+
+Let's assume an exchange wants to run nodes to identify deposits in the Hathor network, so a recommended approach for the integration would be to run at least two full nodes (node1 and node2), which are not connected between them (node1 must have node2 in the blacklist and vice versa). In that architecture, if any deposit is identified in node1, then the exchange must check that it's also a valid transaction in node2 and in one of the public nodes. With this approach, if one of the nodes is compromised and with a different consensus/state, an alert must be raised, so we must investigate what happened.
```
msbrogli, line 53: I'd change "With this approach, if one of the nodes is compromised and with a different consensus/state, an alert must be raised, so we must investigate what happened" by "In this approach, if an attacker successfully compromises one of your full nodes, your validation would fail and the deposit will not be accepted.".

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Validate all your full nodes have the same best block
+
+Use cases should regularly check whether the best block is the same on all their full nodes. If full node have different best blocks, the validation must be done again some seconds later because this might happen depending on the network block propagation time. If the differente continue, the nodes might be under attack and the use case should consider blocking deposits.
+
+
+## Peer-id
+
+The peer-id is a unique identifier of your full node in Hathor's p2p network. You must keep your peer-id secret to prevent attackers from directly targeting your full nodes. Do not tell anyone your peer-ids, and do not publish them on public channels, only in private groups with the Hathor team. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.
```
msbrogli, line 62: I'd remove "only in private groups with the Hathor team".

text/0000-use-case-integration-best-practices.md
msbrogli: I feel it's weird to have a "correctly validate transactions" inside a "be alert for weird behavior". Maybe move it out?

text/0000-use-case-integration-best-practices.md
msbrogli: I feel it should be a comprehensive document (instead of pointing to other documents). So I feel we should explain what should be checked. For example:

"Version identifies the type of vertex. Here are the choices: ... Depending on your use case, you might have different validation rules for each type."

"Voided state: A void transaction is cancelled and should never be accepted. You must validate that the transaction is not voided asserting that `is_voided == false`.

"Outputs: Hathor supports multi-token transactions. You must confirm that your outputs are correctly placed as follows:

* Token id is correct. If you accept only HTR token, token id must be `"00"`.
* Token data is equal to `0`.
* Value matches the deposit value. Note that it is an integer and `12.34` is represented by `1234`.
* Timelock must be null; otherwise your funds might be locked.
"

text/0000-use-case-integration-best-practices.md
msbrogli, line 108: I'd say "won't become voided in the future"

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,175 @@
...
+
+Some use cases might handle transactions with huge amounts, so it's essential to wait for some blocks to confirm the transaction before accepting it as a valid one. The more blocks confirm a transaction, the more guarantee there is that this transaction won't become invalid in the future. As a reference, Bitcoin's use cases usually require a six confirmations before accepting a new deposit.
+
+### Check if an unusual amount of deposits or withdrawals are being made
+
+Many use cases have withdrawals/deposits in their set of features through blockchain transactions. It's important to check for unusual levels of deposits and withdrawals because this could be caused by an attack.
+
+For that situation, it's crucial to have an easy way to block specific accounts that have unusual behavior and might be part of an attack. You might also consider to limit the number of operations a user can do in a time window.
+
+This validation and user block must be done in the use-case software application.
```
msbrogli, line 116: I'd say "This validation and user throttling must ..."

text/0000-use-case-integration-best-practices.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Check if one of your full nodes gets out-of-sync
+
+One other relevant aspect of full node is to always check for weird behavior is the sync among them. We recommend use cases to regularly validate that all their full nodes are in sync among them and in sync with at least one public node as well.
```
msbrogli, line 120: Something is wrong in this sentence. It seems there's an extra "is" here. A suggestion would be to go straight to the point, for example: "Always check for weird behavior in the synchronization among full nodes."

msbrogli (changes requested):

text/0000-use-case-integration-best-practices.md
pedroferreira1: Changed, it's better now. Thanks

text/0000-use-case-integration-best-practices.md
pedroferreira1: Yeah, better now. And the extra "is" was a typo, it was supposed to be an "in".

text/0000-use-case-integration-best-practices.md
pedroferreira1: Humm maybe. The idea was that any tx with timelock, or from a different token than the expected is a "weird behavior".

I moved the section out anyway, it's good.

text/0000-use-case-integration-best-practices.md
pedroferreira1: I understand your idea but my main concern is that the fields depend on the integration. If the use case is getting information from the headless or the full node, the tx fields might be different.

But the text is generic now, changed to your suggestion. What do you think?

text/0000-use-case-integration-best-practices.md
msbrogli: Great! Thanks!

msbrogli: Yes, maybe we should add a feature to the headless wallet where it confirms a transaction with multiple full nodes.

text/0000-use-case-integration-best-practices.md
msbrogli: I'd completely remove the Hathor team in this sentence. Other than that, it's good to me.

text/0000-use-case-integration-best-practices.md
msbrogli: Great job! I think it's way better. Thanks!

text/0000-use-case-integration-best-practices.md
msbrogli: Thanks!

text/0000-use-case-integration-best-practices.md
msbrogli: Thanks!

msbrogli (approved):

text/0000-use-case-integration-best-practices.md
pedroferreira1: I've removed it. The new text is:

> The peer-id is a unique identifier of your full node in Hathor's p2p network. You must keep your peer-id secret to prevent attackers from directly targeting your full nodes. Do not tell anyone your peer-ids, and do not publish them on public channels. If you think your peer-id has been exposed, you should generate a new peer-id and replace the exposed ones.

text/0000-use-case-integration-best-practices.md
pedroferreira1: Created this issue: https://github.com/HathorNetwork/hathor-wallet-headless/issues/248

PR #47, opened 2023-03-16
feat(events): move reliable integration designs as-is

glevco:
Acceptance Criteria:

- Copy exact current state of issue descriptions for https://github.com/HathorNetwork/hathor-core/issues/405 and https://github.com/HathorNetwork/hathor-core/issues/437 with no changes


jansegre (approved):
Given that this is the de facto design we already accepted before. I'll just approve it and not re-review it.

msbrogli (approved):

PR #50, opened 2023-03-22
docs: headless design for atomic swap service

tuliomir:
View [rendered](https://github.com/HathorNetwork/rfcs/blob/docs/atomic-swap-service-headless/projects/atomic-swap-service/headless-wallet.md)

### DIFF
--- a/projects/atomic-swap-service/headless-wallet.md
+++ b/projects/atomic-swap-service/headless-wallet.md
@@ -1,11 +1,3 @@
-# Table of Contents
-- [Summary](#summary)
-- [Acceptance criteria](#acceptance_criteria)
-- [Guide-level explanation](#guide-level-explanation)
-- [Reference-level explanation](#reference-level-explanation)
-- [Rationale and alternatives](#rationale-and-alternatives)
-- [Future possibilities](#future-possibilities)
-
 # Summary
 [summary]: #summary
 
--- a/projects/atomic-swap-service/headless-wallet.md
+++ b/projects/atomic-swap-service/headless-wallet.md
@@ -4,6 +4,7 @@
 - [Guide-level explanation](#guide-level-explanation)
 - [Reference-level explanation](#reference-level-explanation)
 - [Rationale and alternatives](#rationale-and-alternatives)
+- [Future possibilities](#future-possibilities)
 
 # Summary
 [summary]: #summary
@@ -22,7 +23,7 @@ This will allow for easy swap operations between desktop and headless clients, m
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-The headless wallet, in a similar fashion as what is proposed on the Desktop Wallet, will have a `listnenedProposals` map in memory for each of the initialized wallets and will keep track of their identifiers, passwords and real-time updates.
+The headless wallet, in a similar fashion as [what is proposed on the Desktop Wallet](https://github.com/HathorNetwork/hathor-wallet/pull/361/files#diff-81b333677b8eb7fcc977d225072c1c10453c0aee095f9458885d7a553ef3579d), will have a `listnenedProposals` map in memory for each of the initialized wallets and will keep track of their identifiers, passwords and real-time updates.
 
 ### Example use case 1
 A user for the initialized wallet `swap-website-wallet` calls the `registerProposal` endpoint with an identifier `prop-1` and password `123` that it has received from another one of the proposal's participant. This request will add `prop-1` to the local `listenedProposals` object in memory.
@@ -46,7 +47,7 @@ There are some considerations about this object:
   - There should be no more than one websocket connection open for each proposal
 - Wallets that did not register/create a proposal should not receive updates about it
 
-To achieve this, a `listenedProposals` map will be created on the `services/wallets.service` module containing, in short:
+To achieve this, a `listenedProposals` map will be created on the `services/wallets.service` module containing, in short, the proposal identifier, password, and a list of wallet-ids that are listening to it:
 ```ts
 {
 	proposalId: string,
@@ -88,8 +89,13 @@ The [current workflow](https://hathor.gitbook.io/hathor/guides/headless-wallet/a
 
 Calling this route with a proposal identifier that has not been registered will raise a `400` error. Any errors raised while interacting with the service will also be treated and returned on the http response.
 
+### Unregister
+A new route `[DELETE] /wallet/atomic-swap/tx-proposal/delete/{proposalId}` will be created, allowing the user to stop listening to this proposal for the wallet specified on the request header.
+
+Should this wallet be the last one listening to the proposal, it should also be removed from the map and its websocket connection closed.
+
 ### External notifications
-For every listened proposal a web-socket connection will be opened with the Atomic Swap Service, informing the application about any changes that happen to it through the event `wallet:update-proposal`.
+For every listened proposal a websocket connection will be opened with the Atomic Swap Service, informing the application about any changes that happen to it through the event `wallet:update-proposal`.
 
 Whenever a message arrive through this channel, all wallets that listen to the updated proposal will receive a notification through the _External Notifications_ feature containing the updated proposal contents.
 
@@ -97,11 +103,11 @@ Whenever a message arrive through this channel, all wallets that listen to the u
 [rationale-and-alternatives]: #rationale-and-alternatives
 
 ### Lean proposal objects
-In contrast to the desktop wallet, that keeps all of the proposal data in memory, the headless keeps only the identifier, password and listeners.
+In contrast to the desktop wallet, that keeps all the proposal data in memory, the headless keeps only the identifier, password and listeners.
 
 The current atomic swap headless workflow already expects the application not to keep any of the proposal data cached, so this would not present a difference in its usability.
 
-The alternative of keeping all of the proposal's serialized `PartialTx` and `signatures` would not add any relevant ease of use while increasing the application's memory consumption.
+The alternative of keeping all the proposal's serialized `PartialTx` and `signatures` would not add any relevant ease of use while increasing the application's memory consumption.
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
--- a/projects/atomic-swap-service/headless-wallet.md
+++ b/projects/atomic-swap-service/headless-wallet.md
@@ -34,23 +34,37 @@ These identifier and password can then be shared with the other interested parti
 This integration with the swap service backend needs a way for the headless to store in memory the proposals that are currently being monitored by each of its initialized wallets.
 
 There are some considerations about this object:
-- An event listener will be created for each proposal to receive web-socket updates from the service
+- An event listener will be created for each wallet to receive web-socket updates from the service
 - More than one initialized wallet can be listening to the same proposal
-  - There should be no more than one websocket connection open for each proposal
-- Wallets that did not register/create a proposal should not receive updates about it
+- Wallets that did not register a proposal should not receive updates about it
 
-To achieve this, a `listenedProposals` map will be created on the `services/wallets.service` module containing, in short, the proposal identifier, password, and a list of wallet-ids that are listening to it:
-```ts
-{
-	proposalId: string,
-	password: string,
-	wallets: string[],
-}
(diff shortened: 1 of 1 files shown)


projects/atomic-swap-service/headless-wallet.md
```diff
@@ -0,0 +1,118 @@
...
+- Wallets that did not register/create a proposal should not receive updates about it
+
+To achieve this, a `listenedProposals` map will be created on the `services/wallets.service` module containing, in short, the proposal identifier, password, and a list of wallet-ids that are listening to it:
+```ts
+{
+	proposalId: string,
+	password: string,
+	wallets: string[],
+}
+```
```
r4mmer, line 57:
```ts
interface txProposalConfig {
  id: string,
  password: string,
}

// keys are proposal ids
type walletProposals = Map<string, txProposalConfig>;

// keys are wallet-ids
const proposals = new Map<string, walletProposals>();
```

We don't need to create the intermediate type, I just used it for clarification.
The idea is that if we need to get all the proposals this wallet is listening this would be faster.
And we will need this operation since we may want to stop listening to all proposals of a wallet when the wallet is stopped.

r4mmer: BTW, I think we need an api to list current proposals we are listening.
For instance, if a user wants to unregister a proposal but lost the id, he could list the proposals and unregister.

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -0,0 +1,118 @@
+# Table of Contents
+- [Summary](#summary)
+- [Acceptance criteria](#acceptance_criteria)
+- [Guide-level explanation](#guide-level-explanation)
+- [Reference-level explanation](#reference-level-explanation)
+- [Rationale and alternatives](#rationale-and-alternatives)
+- [Future possibilities](#future-possibilities)
```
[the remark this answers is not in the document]
tuliomir: I had to activate the `Feature Preview` option on GitHub to see this option, it wasn't available before. It's still unavailable for incognito browsers or such.

Since the main use case for this document is to be read by us on GitHub, I think we don't need it anymore. Removed.

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -0,0 +1,118 @@
...
+- An event listener will be created for each proposal to receive web-socket updates from the service
+- More than one initialized wallet can be listening to the same proposal
+  - There should be no more than one websocket connection open for each proposal
+- Wallets that did not register/create a proposal should not receive updates about it
+
+To achieve this, a `listenedProposals` map will be created on the `services/wallets.service` module containing, in short, the proposal identifier, password, and a list of wallet-ids that are listening to it:
+```ts
+{
+	proposalId: string,
+	password: string,
```
pedroferreira1, line 54: Do you need to keep the `password`, or can you just store the `hash` that is used to call the service APIs?

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -0,0 +1,118 @@
...
+These identifier and password can then be shared with the other interested participants and future interactions can follow just like on example 1.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Listened proposals
+This integration with the swap service backend needs a way for the headless to store in memory the proposals that are currently being monitored by each of its initialized wallets.
+
+There are some considerations about this object:
+- An event listener will be created for each proposal to receive web-socket updates from the service
```
pedroferreira1, line 45: Why do you need one ws connection for each proposal? Imagine a use case that handles many atomic swaps. Given that we don't know when an atomic swap is completed, we would keep the ws connection opened until the user restarts the headless or unregister it.

I think the best way to protect this is to have a single ws connection and we receive the proposal id in the ws message. What do you think?

projects/atomic-swap-service/headless-wallet.md
tuliomir: Yes, the lib needs the original `password` to decrypt the API response contents.

We could encrypt it locally somehow, but that would require another kind of password. Since we only store this information in memory, I think it's not worth it for now.

projects/atomic-swap-service/headless-wallet.md
tuliomir: Good suggestion, changed the document to add it. Also added a route for the list of `listenedProposals`

projects/atomic-swap-service/headless-wallet.md
tuliomir: That's a good point, changed the document to reflect it.

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -42,23 +34,37 @@ These identifier and password can then be shared with the other interested parti
...
+ * @type {Record<string,WalletListenedProposals>}
+ */
+const walletListenedProposals = {};
 ```
 
-Should any request for this proposal return an unrecoverable error, for example a `404` response that indicates the proposal has already expired, it should be automatically removed from memory.
+For simplicity, references to `listenedProposals` on this document refer to the listened proposals for a specific initialized wallet
 
-### Create
+Should any request for a proposal return an unrecoverable error, for example a `404` response that indicates the proposal has already expired, it should be automatically removed from memory.
```
r4mmer, line 65: How are we planning to do this?
Since only some routes will have the proposalId we could have a middleware to load the proposal from the `listenedProposals` and if an error is raised, e.g. `AtomicSwapError` or `ProposalUnrecoverableFaillure`, we can have the cleanup process on the middleware instead of on each method.

The downside of this is that some routes expect the proposal id as a body argument and others as a route parameter.
I think there are 2 alternatives:

1. We could have the `routes` function extract the proposalId and pass it to the middleware.
2. Move all routes that use proposalId to receive in the same way, e.g. `X-PROPOSAL-ID` header or on the querystring

The option 1. would be something along the lines of:

```javascript
router.get(
  '/register/:proposalId',
  registerSchema,
  async (req, res) => {
    return atomicSwapProposalWrapper(req.params.proposalId, registerController)(req, res);
  },
);
```

And the second can be a normal middleware, but at the cost of flexibility on the api

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -68,7 +74,7 @@ The `[POST] /wallet/atomic-swap/tx-proposal` route will receive the new optional
 ```
 Upon a successful creation, a new property `createdProposalId` will be available on the response object, and its data will be registered on the `listenedProposals` accordingly.
 
-### Register
+### Register endpoint
 A new route `[POST] /wallet/atomic-swap-tx-proposal/register/${proposalId}` will be created, receiving as body parameters:
```
r4mmer, line 78:
```suggestion
A new route `[POST] /wallet/atomic-swap/tx-proposal/register/${proposalId}` will be created, receiving as body parameters:
```

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -79,25 +85,28 @@ This will allow the wallet user to add any existing proposal to the wallet's `li
 
 To ensure the informed parameters are correct, a `get` request will also be made at this time to the service backend, and a validation executed on the received contents. Any failure found on the process will be thrown and the registration discarded.
 
-### Fetch
+### Listened Proposals endpoint
+A new route `[GET] /wallet/atomic-swap/tx-proposal/listenedProposals` will be created, allowing the user to retrieve the list of `proposalIds` currently being listened to. No password should be retrieved on this call.
```
r4mmer, line 89: Maybe `/wallet/atomic-swap/tx-proposal/list` ?
What do you think?

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -79,25 +85,28 @@ This will allow the wallet user to add any existing proposal to the wallet's `li
...
 
-Should this wallet be the last one listening to the proposal, it should also be removed from the map and its websocket connection closed.
+Also, should a wallet be stopped, all proposals it's currently listening to should also be removed, and its websocket connection closed.
 
 ### External notifications
-For every listened proposal a websocket connection will be opened with the Atomic Swap Service, informing the application about any changes that happen to it through the event `wallet:update-proposal`.
+For every wallet with a registered proposal, a websocket connection will be opened with the Atomic Swap Service, informing the application about any changes that happen to its proposals through the event `wallet:update-proposal`.
 
-Whenever a message arrive through this channel, all wallets that listen to the updated proposal will receive a notification through the _External Notifications_ feature containing the updated proposal contents.
+Whenever a message arrive through this channel, the corresponding wallet will receive a notification through the _External Notifications_ feature containing the updated proposal full contents.
```
r4mmer, line 109: Do we know what are the events that will be received?
We should have at least a list of them, for documentation and to check if the names won't collide with existing events on the external notifications.

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -71,20 +85,23 @@ This will allow the wallet user to add any existing proposal to the wallet's `li
 
 To ensure the informed parameters are correct, a `get` request will also be made at this time to the service backend, and a validation executed on the received contents. Any failure found on the process will be thrown and the registration discarded.
 
-### Fetch
+### Listened Proposals endpoint
+A new route `[GET] /wallet/atomic-swap/tx-proposal/listenedProposals` will be created, allowing the user to retrieve the list of `proposalIds` currently being listened to. No password should be retrieved on this call.
```
pedroferreira1, line 89: I would call it `registered-proposals`, because you used `register` in the other API

projects/atomic-swap-service/headless-wallet.md
```diff
@@ -104,9 +104,9 @@ A new route `[DELETE] /wallet/atomic-swap/tx-proposal/delete/{proposalId}` will
 Also, should a wallet be stopped, all proposals it's currently listening to should also be removed, and its websocket connection closed.
 
 ### External notifications
-For every listened proposal a websocket connection will be opened with the Atomic Swap Service, informing the application about any changes that happen to it through the event `wallet:update-proposal`.
+For every wallet with a registered proposal, a websocket connection will be opened with the Atomic Swap Service, informing the application about any changes that happen to its proposals through the event `wallet:update-proposal`.
```
pedroferreira1, line 107: I don't really like the name `wallet:update-proposal`. A proposal can reference a lot of things, in this case an atomic swap proposal.

I would use `wallet:update-atomic-swap-proposal`

projects/atomic-swap-service/headless-wallet.md
tuliomir: After a more careful listing of all possible failures, I concluded the only irrecoverable one is the `404` from the backend on the `register`, `get` and `update` routes. 
Besides that, only on the `register`, the `403` error indicates the user has actually informed wrong inputs for the registration.

This is because the `register` endpoint already guarantees that any proposal inside it will contain the correct password, so no further validation is required at the other endpoints.

Because of that, I believe having a middleware would not be necessary, as the controllers themselves can have the necessary logic without becoming bloated.

Do you agree?

projects/atomic-swap-service/headless-wallet.md
tuliomir: Fixed!

projects/atomic-swap-service/headless-wallet.md
tuliomir: It's a good suggestion: renamed it.

projects/atomic-swap-service/headless-wallet.md
tuliomir: Good point: added a list of all the events, both from the wallet lib and the ones emitted on the external notifications.

projects/atomic-swap-service/headless-wallet.md
tuliomir: Changed it to `/list` , based on a suggestion from a comment above.

projects/atomic-swap-service/headless-wallet.md
tuliomir: Good point! Added the description with the names of the events both from the wallet lib and the one emitted at the external notifications.

r4mmer (approved):

PR #51, opened 2023-04-05
Feature Activation design

glevco:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/feat/feature-activation/projects/feature-activation/0001-feature-activation-for-blocks.md)


projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+
+It is allowed to change code that is conditional on feature activation, but **NOT** to remove a feature from the definition, as described in the [Feature and Criteria definition] section.
+
+## Feature and Criteria definition
+[Feature and Criteria definition]: #feature-and-criteria-definition
+
+The definition of new features available for feature activation should be as simple as defining the feature name and the criteria required to evaluate its activation state. Once defined, ideally, it should not be changed. The `Feature` enum will contain all features, past and future, activated or not:
+
+```python
+class Feature(Enum):
```
msbrogli, line 120: I feel we should use a separate class for each feature. It would allow each feature to be described and configured in a separate file. Another benefit would be to add custom behavior in the future. What do you think? I also think we should use a separate class per network.

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+    NANO_CONTRACTS = Criteria(name='NANO_CONTRACTS', ...)
+    SOME_NEW_OP_CODE = Criteria(name='SOME_NEW_OP_CODE', ...)
+    SOME_OTHER_FEATURE = Criteria(name='SOME_OTHER_FEATURE', ...)
+```
+
+The `Criteria` data class is used to define attributes such as the feature name and activation requirements. Each name should be unique across all features, past and future. Its definition is provided in the [Criteria configuration] section.
+
+Different features with different criteria should be configurable separately for each network, mainnet and testnet. Feature definitions should **NEVER** be removed, even after activation, as explained in the [Burying feature activations] section. This is meant to preserve history.
```
msbrogli, line 128: How would we pick which features are available for each network?

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+[Feature States]: #feature-states
+
+For each block, there's a state associated with each feature. Possible states are similar to BIP 8, but `LOCKED_IN` and `MUST_SIGNAL` are left out and kept as future possibilities. These are the possible states:
+
+- `DEFINED`: Represents that a feature is defined. It's the first state for each feature. The genesis block is by definition in this state for all features.
+- `STARTED`: Represents that the activation process for some feature is started. Blocks at or above the start height are in this state.
+- `ACTIVE`: Represents that a certain feature is activated. Blocks are in this state if the evaluation criteria were met in the previous evaluation interval.
+- `FAILED`: Represents that a certain feature is not and will never be activated. Blocks are in this state if they're above the timeout height and `activate_on_timeout` is `false` (more on that below).
+
+Note: features in the `FAILED` state will indeed never be activated, but that doesn't mean a new activation process for the same feature couldn't be retried with a different definition.
```
msbrogli, line 140: Maybe add a note that, after resolution, the mining bit would be available for another feature.

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+
+### State retrieval
+[State retrieval]: #state-retrieval
+
+Feature activation will be implemented as a method in `BaseTransaction` that users (that is, developers implementing features in `hathor-core`) can use to conditionally execute different code based on the activation criteria for a certain feature.
```
msbrogli, line 145: Shouldn't we place it inside the consensus algorithm or the best block? Maybe we need another central entity to manage it.

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+The `Criteria` data class is used to define attributes such as the feature name and activation requirements. Each name should be unique across all features, past and future. Its definition is provided in the [Criteria configuration] section.
+
+Different features with different criteria should be configurable separately for each network, mainnet and testnet. Feature definitions should **NEVER** be removed, even after activation, as explained in the [Burying feature activations] section. This is meant to preserve history.
+
+## Feature States
+[Feature States]: #feature-states
+
+For each block, there's a state associated with each feature. Possible states are similar to BIP 8, but `LOCKED_IN` and `MUST_SIGNAL` are left out and kept as future possibilities. These are the possible states:
+
+- `DEFINED`: Represents that a feature is defined. It's the first state for each feature. The genesis block is by definition in this state for all features.
```
msbrogli, line 135: What happens if a fork happens and get back to the start height? Would the feature be evaluated again? Should we keep track of the feature state per block?

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+
+> The version field takes up 2 bytes. The leftmost byte is reserved for future use and ignored for now. This means that a version with value [0xAA, 0x01] should be interpreted just as `version=1`.
+
+I propose that only the 4 rightmost bits of the leftmost byte are used for feature flag bit signaling. That would still keep 4 bits reserved for future use and ignored, while allowing for up to 4 simultaneous features in different activation processes. If necessary, in the future those 4 leftmost bits could be used to change the interpretation of the 4 rightmost bits. Some examples:
+
+- `[0xA, 0x0, 0x01]`: represents [ignored, no feature support, `version=1`]
+- `[0xC, 0x1, 0x02]`: represents [ignored, feature support for `bit=0`, `version=2`]
+- `[0x0, 0x5, 0x03]`: represents [ignored, feature support for `bit=0` and `bit=2`, `version=3`]
+
+Note: if a miner sends a bit signal for a bit that is not defined as a feature bit, it is going to be ignored.
```
msbrogli, line 229: For clarity, to be ignored means that the block will be accepted, right?

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+
+## Criteria configuration
+[Criteria configuration]: #criteria-configuration
+
+The `Criteria` data class represents the configuration for a certain feature activation criteria:
+
+```python
+class Criteria(BaseModel):
+    name: str
+    bit: int
```
msbrogli, line 247: Add a comment explaining whether it is LSB or MSB.

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+#### `name`
+
+A string representing the name of the feature. Must be unique across all features, past and future.
+
+#### `bit`
+[bit]: #bit
+
+Represents which bit in the `version` field of the block is going to be used to signal the feature support by miners. Must be chosen from the set {0,1,2,3}, as defined in the [Bit signaling implementation] section.
+
+Must be selected such that no two concurrent features use the same bit. Also, a bit should only be reused if the `start_height` of a new feature is greater than the `timeout_height` of the previous feature that used this bit.
```
msbrogli, line 267: Add an example for how to select the bit (ex: `version[0] & (1 << bit)`).

msbrogli (changes requested):

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+
+Specifies the minimum number of blocks per evaluation interval required to activate the feature. Usually calculated from a percentage of `EVALUATION_INTERVAL`.
+
+Here it is defined as 36288 blocks (90% of `EVALUATION_INTERVAL`) for the mainnet and 30240 blocks (75% of `EVALUATION_INTERVAL`) for the testnet, as default values. It may be different for each feature, though.
+
+#### `minimum_activation_height`
+
+Specifies the height of the first block at which the feature is allowed to become active. If it's smaller than `start_height` (usually 0), then there's no restriction on when this feature can become active.
+
+Must be less than or equal to `timeout_height`. Must be an exact multiple of `EVALUATION_INTERVAL` (that is, at an evaluation interval boundary).
```
msbrogli, line 291: Should we create a `validate()` method for the feature flags? It seems we can perform some verifications in the CI or as part of the release process.

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+```
+
+## Feature Service
+[Feature Service]: #feature-service
+
+As explained in the [State retrieval] section, a `BaseTransaction` will contain a `feature_service` property of the `FeatureService` class. That class is responsible for handling all logic related to feature activation. There are three methods exposed by the feature activation service:
+
+```python
+class FeatureService:
+    def is_feature_active(self, vertex: BaseTransaction, feature: Feature) -> bool:
```
msbrogli, line 339: Should it receive a Block instead?

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+
+Therefore, it should be enough to store this cache in memory. It would need to be reconstructed every time the full node is restarted. If during implementation tests this reconstruction is deemed to be too slow, RocksDB may be used to permanently store it.
+
+### Dealing with reorgs
+
+When a reorg happens, re-computation of states may be necessary. The cache must be voided for every cached block that participated in the reorg.
+
+### Mempool
+
+Transactions in the mempool should behave naturally by getting the state of its respective closest parent block.
```
msbrogli, line 447: For clarity, wouldn't the closest parent block be the best block?

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+
+The state of a block/feature combination is completely determined by the state of the first block in an evaluation period, that is, the previous block with a height that is a multiple of `EVALUATION_INTERVAL`. By caching every one of those interval boundary blocks for each feature, the problem is resolved.
+
+Given the defined `EVALUATION_INTERVAL` value of `40320`, and the current blockchain height of approximately 3,400,000 blocks, the cache will store less than 100 boundary block states for each feature, and that number will be small for years.
+
+Therefore, it should be enough to store this cache in memory. It would need to be reconstructed every time the full node is restarted. If during implementation tests this reconstruction is deemed to be too slow, RocksDB may be used to permanently store it.
+
+### Dealing with reorgs
+
+When a reorg happens, re-computation of states may be necessary. The cache must be voided for every cached block that participated in the reorg.
```
msbrogli, line 443: I missed more details about how reorgs should be handled. I'm assuming that each blockchain would have its own set of features. In order words, even if a given feature was activated, a reorg might disable that feature (and vice-versa).

projects/feature-activation/0001-feature-activation.md
```diff
@@ -0,0 +1,558 @@
...
+  FAILED --> FAILED
+```
+
+## Feature Service
+[Feature Service]: #feature-service
+
+As explained in the [State retrieval] section, a `BaseTransaction` will contain a `feature_service` property of the `FeatureService` class. That class is responsible for handling all logic related to feature activation. There are three methods exposed by the feature activation service:
+
+```python
+class FeatureService:
```
msbrogli, line 338: Maybe add a `bits_description()` method that would return the feature linked to each bit?

msbrogli (changes requested):

projects/feature-activation/0001-feature-activation.md
glevco: Makes sense. Maybe the already specified `Criteria` class serves this purpose, and the change would be just on the `Feature` enum.

Each feature could have its separate file, just defining its `Criteria` instance as a constant. in a `features/nano_contracts.py` file:

```python
NANO_CONTRACTS = Criteria(name='NANO_CONTRACTS', ...)
```

Then, in a `features.py` file, there could be a list of features:

```python
FEATURES: List[Criteria] = [
    NANO_CONTRACTS,
    ...
]
```

Does that work?

projects/feature-activation/0001-feature-activation.md
glevco: What do you mean by pick? Like, how the code picks it or how humans would select which feature to specify?

projects/feature-activation/0001-feature-activation.md
glevco: Done in https://github.com/HathorNetwork/rfcs/pull/51/commits/6a3adcb62aa2b5bc7c0802c9f363f840da6f70a8

projects/feature-activation/0001-feature-activation.md
glevco: As explained in the Feature Service section, all code related to feature activation would be in a separate class, the `FeatureService`, that may be that central entity. A `BaseTransaction` method is provided just for convenience, and its only purpose is to call the `FeatureService`.

By doing that, feature activation logic is automatically defined for the best block, as it is a `BaseTransaction`. All blocks can access the `FeatureService` and query whether some feature is active or not, up to that block itself.

It's also available on the consensus via the same `BaseTransaction` method, for both blocks and transactions (as both are `BaseTransaction`s. Feature states cannot be queried without a block, as block information is necessary to determine the feature state. If at some point the consensus requires to query a feature state in a moment it's not handling a block, it should first get the best block and then query for the feature state using the best block access to the `FeatureService`.

Please let me know if that covers the use cases you're thinking of.

projects/feature-activation/0001-feature-activation.md
glevco: Every feature is evaluated dynamically every time its state is queried on a specific block, by checking the state of that feature on all blocks up to the genesis. To make this efficient, every N blocks the state is cached, where N is the evaluation period time interval, defined in number of blocks.

When forks happen, the cache must be invalidated for every voided block. By doing that, the cache is preserved for blocks that weren't changed, and the new feature state is recalculated from the common block downwards.

Reference sections for this are "Caching mechanism", "Dealing with reorgs", and "Example".

projects/feature-activation/0001-feature-activation.md
glevco: In the sense that it is a valid bit configuration, yes, the block will be accepted. I've updated the note.

But also for clarity, we may define some new feature that accepts or rejects blocks based on some condition. Then, if that feature is in the `ACTIVE` state, whether the block will be accepted or rejected depends on that condition.

projects/feature-activation/0001-feature-activation.md
glevco: Added in the "Bit signaling implementation" section, in https://github.com/HathorNetwork/rfcs/pull/51/commits/6a3adcb62aa2b5bc7c0802c9f363f840da6f70a8

projects/feature-activation/0001-feature-activation.md
glevco: Added in the "Bit signaling implementation" section, in https://github.com/HathorNetwork/rfcs/pull/51/commits/6a3adcb62aa2b5bc7c0802c9f363f840da6f70a8

projects/feature-activation/0001-feature-activation.md
glevco: Ok, I'll add it to the design. Do you have an idea on where it could be used?

projects/feature-activation/0001-feature-activation.md
glevco: I'm not sure. That's probably a knowledge gap for me. Is the best block always the closest parent block for all transactions in the mempool?

projects/feature-activation/0001-feature-activation.md
glevco: Makes sense. Internal rules will be validated by the `Criteria` class itself, for example, checking that `minimum_activation_height` is less than or equal to `timeout_height` and a multiple of `EVALUATION_INTERVAL`.

That `validate()` method would be useful to check for rules between different features, like checking that no two features are using the same name, or the same bit concurrently. I'll add it to the design.

projects/feature-activation/0001-feature-activation.md
glevco: Indeed. I added some more explanation on https://github.com/HathorNetwork/rfcs/pull/51/commits/b43a33cd34037aa3cc625c5b77ae99b407f9fad3. Let me know if you're still missing more details.

However, I do believe there's some low-level ambiguity left open here, that may only be clearer during implementation. I've added one more dev day to the reorg task to factor that in.

projects/feature-activation/0001-feature-activation.md
glevco: Done in https://github.com/HathorNetwork/rfcs/pull/51/commits/1230bbad8164aaf441b9ba70009b84f734f7d973

projects/feature-activation/0001-feature-activation.md
msbrogli: It does! Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: I was asking how each network would list its available features. But, as each feature would be for only one network, it does not seem to be a problem. Make sense?

projects/feature-activation/0001-feature-activation.md
msbrogli: Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: It does. I'm ok moving forward as you explained. We can discuss the specifics during the implementation review.

projects/feature-activation/0001-feature-activation.md
msbrogli: Sure. Thanks for the clarification.

projects/feature-activation/0001-feature-activation.md
msbrogli: Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: Great! Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: I feel that we should change the explorer to show these bits for each block. It would include showing the bit number, the feature associated to that bit, and the state of the feature. What do you think? If you agree, let's add it to the list of tasks for this project.

projects/feature-activation/0001-feature-activation.md
msbrogli: Got it. I didn't understand how the cache would work. Maybe I should read this section again. One suggestion might be to store information in blocks metadata. For instance, you might keep a dict of features and states. In this case, each block would copy this metadata from its parent and update according to its own bits. What do you think?

projects/feature-activation/0001-feature-activation.md
msbrogli: It seems likely that we will have features that affect transactions. If I got it right, you will add a helper to transactions so it would get the features from the closest block. In this case, isn't it better to add a method `BaseTransaction.get_feature_service()` and implement something like `return FeatureService(self.get_first_block())`?

projects/feature-activation/0001-feature-activation.md
msbrogli: I'm not sure what you are calling as the closest parent block for a transaction in the mempool. The best block is the latest block that belongs to the best blockchain. The best blockchain is the winner blockchains (and all other chains are voided).

projects/feature-activation/0001-feature-activation.md
msbrogli: Good explanation. Thanks! I think that a simpler implementation would use metadata to store information for features, so you don't have the trouble to handle with this maybe-too-complex cache policy.

projects/feature-activation/0001-feature-activation.md
glevco: Actually, after revisiting this, I suggest the `FEATURES` list is actually a list inside `HathorSettings`. The individual file with a `Criteria` instance for each feature remains. But the feature list would be defined in the settings. This enforces a specific list is used by the `FeatureService`.

projects/feature-activation/0001-feature-activation.md
glevco: Oh I see, makes sense. Indeed, that wasn't mentioned. As just commented [here](https://github.com/HathorNetwork/rfcs/pull/51#discussion_r1180764556), the list of features under the activation process would be defined in `HathorSettings`, for each network.

projects/feature-activation/0001-feature-activation.md
glevco: I agree, I'll add it!

projects/feature-activation/0001-feature-activation.md
glevco: Done in https://github.com/HathorNetwork/rfcs/pull/51/commits/7b70d95ec05dc13a16f73f4e6fef31991c960525

projects/feature-activation/0001-feature-activation.md
glevco: Done in https://github.com/HathorNetwork/rfcs/pull/51/commits/7b70d95ec05dc13a16f73f4e6fef31991c960525

projects/feature-activation/0001-feature-activation.md
glevco: That's a great suggestion, using a new metadata attribute to store states prevents the need of creating a new caching structure. But anyway, the caching logic remains mostly the same: when to set, retrieve, and remove the states from the metadata attribute.

Updated in https://github.com/HathorNetwork/rfcs/pull/51/commits/1c874a535200492260e8b5deb5c837df61629c11

projects/feature-activation/0001-feature-activation.md
glevco: Indeed it'll be more or less like you described it. But as agreed, feature activation for transactions will be moved to a separate document. For this reason, I changed the `FeatureService` API to only support `Block`s, as you suggested.

Updated in https://github.com/HathorNetwork/rfcs/pull/51/commits/5b4878b3bcfbe2260a607ec8b169f41dab31305e

projects/feature-activation/0001-feature-activation.md
glevco: Updated as explained in https://github.com/HathorNetwork/rfcs/pull/51#discussion_r1183923947

glevco:
I have updated the design according to the threads and recent meeting notes, including updating the task list and dev days estimation.

projects/feature-activation/0001-feature-activation.md
msbrogli: Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: Great! Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: Great! Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: Ok. Thanks!

projects/feature-activation/0001-feature-activation.md
msbrogli: Ok. Thanks!

projects/feature-activation/0001-feature-activation.md
glevco: @msbrogli I'm unresolving this discussion, as I'm now implementing this and rethinking it again.

As we're currently implementing settings via yaml in other PRs, I believe each network's yaml file would be the best place to define features for feature activation. So in the yaml we would have something like:

```yaml
# mainnet.yml

features:
  - name: NANO_CONTRACTS
    bit: 0
    start_height: 1000
    threshold: 900
    ...
  - name: MY_OTHER_FEATURE:
    bit: 1
    start_height: 1000
    threshold: 900
    ...
```

Previously, you said:

> I feel we should use a separate class for each feature. It would allow each feature to be described and configured in a separate file.

This solution covers this requirement, as each network has its own separate yaml file.

> Another benefit would be to add custom behavior in the future.

Could you please elaborate on this? What do you mean by custom behavior? Maybe this would be covered by adding another attribute into the feature yaml definiton. Custom behavior would be configured via settings.

> I also think we should use a separate class per network.

I believe everything that is variable by network should be defined in settings, not in code.

Another advantage I see in doing this is that we have a single point where you can see all features defined at once. This helps selecting parameters for a new feature, as you can see for example what bits are being used. What do you think?

glevco: This is also related to your other comment [here](https://github.com/HathorNetwork/rfcs/commit/7b70d95ec05dc13a16f73f4e6fef31991c960525#r112651756). Yes, each feature definiton works like an "internal object", but is actually just a dictionary in the yaml file. A `NamedTuple` is used to enforce the schema.

I think this case is different than you concern since in the yaml this is simply a dict with scalar values, contrary to an internal object actually being instantiated in python code.

msbrogli: Hi @glevco . I initially had some mixed feelings about your suggestion. But, after some thinking, I agree that it is a nice approach. If I got it right, each feature will have its own class (e.g., `NanoContractsFeature`) and a manager would load the classes according to the settings file. Is that right?

On custom behavior, I was thinking that some features might have other methods and attributes that would be useful during the execution of the full node. As an example, the `NanoContractFeature` might have a list of the enabled blueprints and it might also carry some attributes used by the consensus algorithm during the execution of a contract. I mean, these attributes would be part of the feature proposal so they would not be configurable from the settings.

glevco: > If I got it right, each feature will have its own class (e.g., NanoContractsFeature) and a manager would load the classes according to the settings file. Is that right?

Not exactly. Let's discuss it over a call, I think there's some misunderstanding on the configuration of features

msbrogli: As discussed in a meeting, I agree that each feature will be just a boolean flag.

jansegre (approved):

jansegre:
I like the proposed design. The design itself is approved for me. After the latest changes maybe we should call it `0001-feature-activation-for-blocks` or something to that effect. What do you think?

glevco:
Thanks, @jansegre. Makes sense, I had forgotten to change the filename. Done in https://github.com/HathorNetwork/rfcs/pull/51/commits/40d32d84f05ed996637e726865ae652629a1dd48!

msbrogli (approved):

PR #48, opened 2023-03-17
feat(events): update reliable integration designs

glevco:
Acceptance Criteria:

- Update high-level design
- Update low-level design
- Aspects related to handling events during the load phase will be on a separate PR

PR #52, opened 2023-04-06
feat(events): realiable integration during the load phase 

glevco:
Depends on https://github.com/HathorNetwork/rfcs/pull/48

Acceptance Criteria:

- Remove the `--emit-load-events` flag since it's not necessary given this design update
- Move all text related to load phase from the high level to the low level design
- Detail how events work during the load phase


projects/reliable-integration/0002-low-level-design.md
```diff
@@ -6,13 +6,49 @@ This low-level design intends to go deeper on the technical aspects of the proje
...
+| `Snapshot start`                  | Some                     | None                   | Empty          | Generate all events, from `event_id = 0`. Events come during both the `LOAD` and `SYNC` phases.                                           |
+| `Restart from aborted LOAD phase` | Some                     | Some                   | `LOAD`         | All events in the database are discarded. Generate all events, from `event_id = 0`. Events come during both the `LOAD` and `SYNC` phases. |
+| `Restart from aborted SYNC phase` | Some                     | Some                   | `SYNC`         | Generate events from `event_id = last_event_id + 1`. All events come during the `SYNC` phase.                                             |
+
+There's no case in which there are some events in the database, but no vertices.
+
+For the `Fresh start` scenario, if we consider there's a `LOAD` phase with 0 events, then the desired outcome becomes the same for the first three scenarios. This makes it possible to simplify the table into only two scenarios, only depending on `State metadata`, as described in the beginning of this section:
```
msbrogli, line 28: Would we emit events for the genesis transactions (two transactions and one block)?

projects/reliable-integration/0002-low-level-design.md
```diff
@@ -6,13 +6,49 @@ This low-level design intends to go deeper on the technical aspects of the proje
...
+| `Restart from aborted SYNC phase`                                      | `SYNC`          | Generate events from `event_id = last_event_id + 1`. All events come during the `SYNC` phase.                                             |
+
+Note: when generating snapshots, events should not be included, as they will be discarded and regenerated anyway.
+
+Now, we can describe a step-by-step representing these rules:
+
+1. The full node begins the initialization process
+2. The `EventManager` starts
+   1. It retrieves the previous node state from the database and stores it
+   2. If its stored previous state is empty or `LOAD`, it discards all events currently in the database
```
msbrogli, line 42: Maybe, after clearing the database, it should add an event of "RESET EVENTS", so an integration can react accordingly to this case. Another point might be to keep the event id instead of resetting it to zero. What do you think?

msbrogli (approved):

projects/reliable-integration/0002-low-level-design.md
jansegre: If we use a UUID for the "event database" and we emit a "event system start" with that UUID, we can have a system that does not have to know internally if it was reset or not. I think that would be better than making the "clearing of a database" something that needs to carefully emit a "reset event". What do you think?

projects/reliable-integration/0002-low-level-design.md
jansegre: I'm in favor of emitting events for genesis vertices on a fresh start instead of having 0 events.

projects/reliable-integration/0002-low-level-design.md
glevco: I agree with @jansegre , and also that UUID could be required to retrieve events. If the client requests events (via HTTP or WebSocket) using the wrong UUID, the full node rejects it, so the the client knows it has to process events from the beginning.

jansegre (approved):

projects/reliable-integration/0002-low-level-design.md
```diff
@@ -6,13 +6,49 @@ This low-level design intends to go deeper on the technical aspects of the proje
...
 ## Full-node execution
 
+Considering this full-node cycle:
+
 ![Full-Node-Cycle drawio](./0002-images/full_node_cycle.png)
 
-### TODO: This section will be updated in a separate PR
+At first, when the full node is initialized, it moves to the `LOAD` state, where data is loaded from the database, without consulting the network. Once all local data is read, the full-node continuously syncs with peers, changing to the `SYNC` state.
+
+In order to know which events will be generated, we will store the states (`LOAD` and `SYNC`) as metadata. So, when the full node starts, we will check the metadata. If it is empty or `LOAD`, we do not know where it stopped and will regenerate all events during load. If it is `SYNC`, we will not generate any events until `LOAD_FINISHED` is emitted.
```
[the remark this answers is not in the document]
glevco: In RocksDB. Added in https://github.com/HathorNetwork/rfcs/pull/52/commits/3a6c667fdbf028613e51def7b455cadf80dec784

projects/reliable-integration/0002-low-level-design.md
glevco: We would. Added in https://github.com/HathorNetwork/rfcs/pull/52/commits/3a6c667fdbf028613e51def7b455cadf80dec784

projects/reliable-integration/0002-low-level-design.md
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

projects/reliable-integration/0002-low-level-design.md
msbrogli: Great! Thanks!

projects/reliable-integration/0002-low-level-design.md
msbrogli: Good approach! So, if I got it right, we will emit a "system start" event, right?

msbrogli (approved):

projects/reliable-integration/0002-low-level-design.md
glevco: Something like it, but it's more an "event system start" than a "system start". It would only be emitted when the event feature is started.

glevco: I'll add this mechanism in a new code PR

PR #54, opened 2023-05-25
feat: web wallet rpc protocol

andreabadesso:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/88305069af54039c67c93f70d1c01e5e6fc9df84/projects/web-wallet/rpc-protocol.md)


projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,330 @@
...
+
+This document outlines the remote procedure call (RPC) methods for a decentralized application (dApp) to interact with Hathor wallets.
+
+The design is currently a minimum viable product (MVP), addressing only known needs from our community. It should be noted that this API design may change as the platform evolves and additional needs arise.
+
+The idea here is to partially replicate the headless wallet API as it is already being used by many partners and has been validated to work on their use cases
+
+# 2. Reference-level explanation
+
+## 2.1 RPC Methods
```
msbrogli, line 11: I missed methods to mint/melt tokens. Such methods might be useful for dApps administrative tasks, for example.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,330 @@
...
+
+We should follow the [JSON-RPC](https://www.jsonrpc.org/specification#error_object) specification for error codes, so errors generated by the "server", which in our case, is the wallet handling the requests should be in the `-32099` to `-32000` range.
+
+| Error | Error Type | Description                        |
+|------------|------------------------------------|---|
+| -32050     | WALLET_NOT_READY | Wallet is not ready yet              |
+| -32051     | REQUEST_REJECTED | Request rejected by the user |
+| -32052     | MISSING_FUNDS | Missing funds to fulfill transaction |
+| -32053 | INVALID_PARAMETERS | Invalid parameters |
+| -32099     | GENERIC_FAILURE | Generic failure |
```
msbrogli, line 270: Maybe UNKNOWN FAILURE or UNEXPECTED FAILURE?

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,330 @@
+# 1. Introduction
+
+This document outlines the remote procedure call (RPC) methods for a decentralized application (dApp) to interact with Hathor wallets.
+
+The design is currently a minimum viable product (MVP), addressing only known needs from our community. It should be noted that this API design may change as the platform evolves and additional needs arise.
+
+The idea here is to partially replicate the headless wallet API as it is already being used by many partners and has been validated to work on their use cases
+
+# 2. Reference-level explanation
```
msbrogli, line 9: I feel this should be a thorough document. So we should add all RPC methods, including the ones already available in the POC.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,330 @@
+# 1. Introduction
```
msbrogli, line 1: Add a link to the Wallet Connect documentation, please. I'm wondering where we validate the network, and how dApps might test their integration using the public testnet or even a private network. Maybe add some words about these topics (or make a reference to the Wallet Connect documentation).

msbrogli (changes requested):

projects/web-wallet/rpc-protocol.md
obiyankenobi: For the initial version, they don't seem very important. I do not see a use case of creating/minting/melting tokens through the web wallet. I think we should have them in the future, but I'm ok having v1 without them.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,330 @@
...
+
+The idea here is to partially replicate the headless wallet API as it is already being used by many partners and has been validated to work on their use cases
+
+# 2. Reference-level explanation
+
+## 2.1 RPC Methods
+
+### 2.1.1 `htr_createTx`
+
+The `htr_createTx` method creates a new transaction based on parameters. It will select the `utxos` automatically, mine and send the transaction
```
obiyankenobi, line 15: Why is this `createTx` and not `sendTx`? Isn't it used to send txs?

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,330 @@
...
+
+**Parameters**
+
+1. `outputs` - An array of outputs
+	1. `address` - Destination address of the output. Required if P2PKH or P2SH
+	2. `value` - (Required if P2PKH or P2SH) The value parameter must be an integer with the value in cents, i.e., 123 means 1.23 HTR.
+	3. `token` - (Defaults to `HTR`) Token id of the output.
+	4. `type` - (Required if data script output) Type of output script.
+	5. `data` - (Required if data script output) Data string of the data script output.
+2. `inputs` - (Optional) Inputs to create the transaction
```
obiyankenobi, line 25: Will we support the filters like in the headless? I mean, so we can specify which addresses to spend from.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,330 @@
...
+This document outlines the remote procedure call (RPC) methods for a decentralized application (dApp) to interact with Hathor wallets.
+
+The design is currently a minimum viable product (MVP), addressing only known needs from our community. It should be noted that this API design may change as the platform evolves and additional needs arise.
+
+The idea here is to partially replicate the headless wallet API as it is already being used by many partners and has been validated to work on their use cases
+
+# 2. Reference-level explanation
+
+## 2.1 RPC Methods
+
```
obiyankenobi, line 12: What about a method to get an address? For instance, the website may want to get the user's address to send some coins.

This would be equivalent to the user copying the address from the wallet and pasting on the website.

tothster:
Within the Dozer perspective, the main methods we will need is based on send and receive funds from/to nano contracts, so I think that this will be very similar(in the parameters and response view) to createTx method already created, but with some points to note:

- Display to the user the contract ID and maybe a link to a explorer when the modal popup at his wallet;
- All methods involved (swap tokens, add liquidity and remove liquidity) will have the input parameter as required ( token to be swapped or LP Token )
- Remove liquidity and swap tokens will trigger nano contract to send funds to user, so maybe the method can have a parameter to create the possibility of user set the address to receive these return funds;

Moreover, with the basic methods already listed, I want to know if it's possible to concatenate them, as we already see on some social logins. For instance, when user first read the QR Code with his wallet, the dApp sends a request to read the balance of some tokens AND also get a address from the wallet, as @obiyankenobi commented above, and this two methods will appear at once on the first modal for user's approval.

projects/web-wallet/rpc-protocol.md
andreabadesso: Yes! Done in 5e1d1efb147d1c61347842219bfb542f61a39f64

projects/web-wallet/rpc-protocol.md
andreabadesso: Added the link to the wallet-connect RFC and to the Web3Wallet docs in 8b170640a819a1888546d2b0960201ef695bf01a

I also added a new RPC, to get the wallet's current connected network in c53a287158be9f89d99f8c60bffec59cddc57693

What do you think?

projects/web-wallet/rpc-protocol.md
andreabadesso: It's not only sending transactions, but it's also building it according to the parameters

I could rename it to `createAndSendTx`, what do you think?

projects/web-wallet/rpc-protocol.md
andreabadesso: We only have the `signMessage` method in the POC

projects/web-wallet/rpc-protocol.md
andreabadesso: We should, updated the design in e64e5fb7d52b96e35bca5779e852d8d21b7a1227

projects/web-wallet/rpc-protocol.md
andreabadesso: I like the idea, added the RPC method here: 8fd28e7a9bfed60cb7f24b247b03bc057e746b09

The wallet should decide whether to let the user pick an address or return an arbitrary one

projects/web-wallet/rpc-protocol.md
andreabadesso: @obiyankenobi I agree, refactored in 7621950fa0e0af8e089271f13919f9af3f5cf3de

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+These methods are going to be implemented in the wallet, and exposed to dApps using wallet-connect, the design of the integration can be read [here](https://github.com/HathorNetwork/rfcs/blob/master/projects/web-wallet/wallet-connect/0001-design.md) and the wallet-connect design docs can be read [here](https://docs.walletconnect.com/web3wallet/about).
+
+# 2. Reference-level explanation
+
+## 2.1 RPC Methods
+
+### 2.1.1 `htr_sendTx`
+
+The `htr_sendTx` method creates a new transaction based on parameters. It will select the `utxos` automatically, mine and send the transaction
```
msbrogli, line 17: Should we add an optional `push_tx: bool` to allow operations where the transaction will not be pushed by the wallet?

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+    }
+  ],
+  "tokens": []
+}
+```
+
+
+### 2.1.2 `htr_createToken`
+
+The `htr_createToken` method creates a new token transaction based on parameters
```
msbrogli, line 125: Suggestion: `... creates a TokenCreationTransaction which will create a new token in the blockchain.`

msbrogli, line 125: Should we allow the caller to choose the inputs? Maybe a `push_tx: bool` parameter to allow the caller better control?

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+      "index": 0
+    }
+  ]
+}
+```
+
+
+### 2.1.4 `htr_signWithAddress`
+
+The `htr_signWithAddress` method requests a signed message using an address' private key that can be verified.
```
msbrogli, line 225: I think we should add that the message will receive a fixed prefix before signing.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+    "message": "sign-me",
+    "address": 5
+  }
+}
+```
+**Response:**
+
+```json
+{
+  "message": "H2/QA48wJhvbKJH6hWXUJPHK5kUTwx4XAxuZGFzUeSPwLeFBv206lOs8ECHMTV/KilGbFP1p5e52mMGTUDFT7WM="
```
msbrogli, line 249: Is it the message followed by the signature? Should we place them in separate fields? Or at least explain how one can split them if needed?

Maybe rename the field to `signed_message`?

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+Different from account-based blockchains where users are used to always using the same address on each transaction, getting the balance on hathor involves going through a list of addresses from the user's wallet and calculating the sum of all utxos.
+
+So in order for a `dApp` to know the balance of a given token for the connected user's wallet, we must have a method to fetch it.
+
+The `htr_getBalance` method fetches the balance for a given token.
+
+**Parameters**
+
+1. `token` - (Defaults to `HTR`) May be specified for a token other than `HTR`
```
msbrogli, line 263: Should we add an optional `addresses` that will return the balance of a subset of addresses? Also maybe a `address_indexes` where the caller can specify the indexes instead of the addresses.

msbrogli, line 263: Should we allow more than one token? So dApps can query for two or three different tokens depending on the operation they are running.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+  "id": 4,
+  "jsonrpc": "2.0",
+  "result": {
+    "available": 500,
+    "locked": 0
+  }
+}
+```
+
+### 2.1.6 `htr_getConnectedNetwork`
```
msbrogli, line 290: Should we add a `network` attribute to all queries so any mismatch would cause an error? This can prevent different types of error.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+  "id": 4,
+  "jsonrpc": "2.0",
+  "result": {
+    "network": "mainnet" | "testnet" | "privatenet",
+    "genesisHash": "00000000000000000437c87c...c8bf9e9d2fda7de9763"
+  }
+}
+```
+
+### 2.1.7 `htr_getAddress`
```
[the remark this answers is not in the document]
msbrogli: I'd say we can add three parameters: `index: int`, `first_empty: bool`, and `full_path: str`.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+}
+```
+**Response:**
+
+```json
+{
+  "id": 4,
+  "jsonrpc": "2.0",
+  "result": {
+    "address": "HNnK9wgUVL6Cjzs1K3jpoGgqQTXCqpAnW8"
```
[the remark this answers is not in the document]
msbrogli: I'd say to return both the index and the full path.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+When the request is received, we will validate if the parameters are valid, enforcing the data types and rules described in the Reference-Level explanation
+
+If the validation fails, we will respond with a `INVALID_PARAMETERS` error code
+
+If the validation is successful, we will display a confirmation screen, with all outputs for manual user validation (UX is yet to be discussed). If the user rejects the transaction, we will respond with a `REQUEST_REJECTED` error
+
+If the user accepts the transaction, we will create a `SendTransaction` model (or `SendTransactionWalletService` if on the wallet-service) and execute the send transaction.
+
+Since `JSON-RPC` is designed considering only a single response per request, we will only respond the request when the transaction either failed to be pushed to the fullnode (or the wallet-service) or when it's successfully pushed.
```
msbrogli, line 370: Should we have a `STILL WAITING` return code? In this case, the dApp could get information from a separate `/current-operation-status` API?

projects/web-wallet/rpc-protocol.md
msbrogli, line 249: Return the address, index and full path used to sign the message.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+We might display "checkboxes" before each utxos so the user can choose exactly which ones he want to display
+
+## 3.4 `htr_signWithAddress`
+
+When the request is received, we will validate if the parameters are valid by following the parameter type definitions in the Reference-Level explanation
+
+If the validation fails, we will respond with a `INVALID_PARAMETERS` error code
+
+At this point, we will display the `message` and the `address` to the user, asking for his confirmation. If he rejects the modal, we will respond with an `INVALID_PARAMETERS` error code.
```
msbrogli, line 404: I think we should prefix the message before showing it to the user. If the original message is already prefixed, we can skip it.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+If the validation fails, we will respond with a `INVALID_PARAMETERS` error code
+
+At this point, we will call the `getUtxos` method to retrieve `utxos` that fulfill the requested filters.
+
+Before returning, we will display a confirmation screen (UX yet to be discussed) asking for user confirmation by displaying the `utxos` that will be returned to the `dApp`.
+
+If the user accepts the transaction, we will reply with the array of `utxos`. Otherwise, we will answer with a `REQUEST_REJECTED` error
+
+#### Future possibilities
```
msbrogli, line 394: I guess the context here is not for future possibilities. Please double check.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+If he accepts, we will call the `signMessageWithAddress` method from the lib and answer the request
+
+## 3.5 `htr_getBalance`
+
+When the request is received, we will validate if the parameters are valid by following the parameter type definitions in the Reference-Level explanation
+
+If the validation fails, we will respond with a `INVALID_PARAMETERS` error code
+
+At this point, we will display the `message` and the `address` to the user, asking for his confirmation. If he rejects the modal, we will respond with an `INVALID_PARAMETERS` error code.
```
msbrogli, line 414: Should we allow users to modify the balance value? We would allow only values smaller than or equal to the current balance.

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+If he accepts, we will call the `getBalance` method from the lib and answer the request
+
+## 3.6 `htr_getConnectedNetwork`
+
+This method should not require explicit approval of the wallet given that the wallet and the dApp are already connected.
+
+It will return the wallet's current connected network, as defined in the reference-level section
```
msbrogli, line 422: Should we notify users that the app is requesting information?

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+## 3.7 `htr_getAddress`
+
+This method is used to request an address from the wallet.
```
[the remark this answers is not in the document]
msbrogli: I think we should let the dApp filter what it wants. If there's more than one choice, than we can give users the ability to choose. What do you think?

msbrogli (changes requested):

projects/web-wallet/rpc-protocol.md
andreabadesso: Refactored it to `UNEXPECTED_FAILURE` in 0b6a49e2be9bb6483ce5494fdf1f1209ba3f4cfb

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+1. `outputs` - An array of outputs
+	1. `address` - Destination address of the output. Required if P2PKH or P2SH
+	2. `value` - (Required if P2PKH or P2SH) The value parameter must be an integer with the value in cents, i.e., 123 means 1.23 HTR.
+	3. `token` - (Defaults to `HTR`) Token id of the output.
+	4. `type` - (Required if data script output) Type of output script.
+	5. `data` - (Required if data script output) Data string of the data script output.
+2. `inputs` - (Optional) An array of inputs to create the transaction
+  1. `type` - Type of input object. Can be either 'query' which is the default or 'specific'
+  1. `hash` - Hash of the transaction being spent in this input. Used if not type query
+  1. `index` - Index of the transction being spent in this input. Used if not type query
```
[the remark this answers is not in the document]
andreabadesso: Thanks fixed in 230a58dcc2fb0fdc46da39a234b439b9f490ba54

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+1. `outputs` - An array of outputs
+	1. `address` - Destination address of the output. Required if P2PKH or P2SH
+	2. `value` - (Required if P2PKH or P2SH) The value parameter must be an integer with the value in cents, i.e., 123 means 1.23 HTR.
+	3. `token` - (Defaults to `HTR`) Token id of the output.
+	4. `type` - (Required if data script output) Type of output script.
+	5. `data` - (Required if data script output) Data string of the data script output.
+2. `inputs` - (Optional) An array of inputs to create the transaction
+  1. `type` - Type of input object. Can be either 'query' which is the default or 'specific'
+  1. `hash` - Hash of the transaction being spent in this input. Used if not type query
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 230a58dcc2fb0fdc46da39a234b439b9f490ba54

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+**Parameters**
+
+1. `outputs` - An array of outputs
+	1. `address` - Destination address of the output. Required if P2PKH or P2SH
+	2. `value` - (Required if P2PKH or P2SH) The value parameter must be an integer with the value in cents, i.e., 123 means 1.23 HTR.
+	3. `token` - (Defaults to `HTR`) Token id of the output.
+	4. `type` - (Required if data script output) Type of output script.
+	5. `data` - (Required if data script output) Data string of the data script output.
+2. `inputs` - (Optional) An array of inputs to create the transaction
+  1. `type` - Type of input object. Can be either 'query' which is the default or 'specific'
```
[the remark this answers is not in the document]
andreabadesso: Fixed in cc5e2a9a934ee4262e5ab9be191e51f7a8e27c16

projects/web-wallet/rpc-protocol.md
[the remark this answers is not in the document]
andreabadesso: Refactored in a5ecf81c53cb4b1546711295f38c05893e853c91, allowing for more types of requests

What do you think?

projects/web-wallet/rpc-protocol.md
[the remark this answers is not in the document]
andreabadesso: What do you think about the refactor in a5ecf81c53cb4b1546711295f38c05893e853c91?

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,432 @@
...
+
+## 3.4 `htr_signWithAddress`
+
+When the request is received, we will validate if the parameters are valid by following the parameter type definitions in the Reference-Level explanation
+
+If the validation fails, we will respond with a `INVALID_PARAMETERS` error code
+
+At this point, we will display the `message` and the `address` to the user, asking for his confirmation. If he rejects the modal, we will respond with an `INVALID_PARAMETERS` error code.
+
+If he accepts, we will call the `signMessageWithAddress` method from the lib and answer the request
```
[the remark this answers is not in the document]
andreabadesso: The lib method doesn't support a custom prefix, but the `bitcore-message` implementation which we use in the lib prefixes the message, we're using `Hathor Signed Message`

You can the bitcore-message implementation [here](https://github.com/bitpay/bitcore-message/blob/98a54319246bfabd442f94cac42ab5c9486d3b46/lib/message.js#L33)

And we're monkey-patching the prefix in our lib [here](https://github.com/HathorNetwork/hathor-wallet-lib/blob/6f37b48cfde35dd4d50054352ecfeb472653d703/src/utils/crypto.ts#L15) as `bitcore-message` doesn't allow us to pass a custom prefix

projects/web-wallet/rpc-protocol.md
andreabadesso: We can, added it in b682c3818f1e2e5f884053518abdbde8668969aa

I also added a new RPC method to push a transaction hex, what do you think?

projects/web-wallet/rpc-protocol.md
andreabadesso: Makes sense, updated in b682c3818f1e2e5f884053518abdbde8668969aa

projects/web-wallet/rpc-protocol.md
andreabadesso: Done in 330bc6059de95fce7a5364b6a8cca197919daa51

projects/web-wallet/rpc-protocol.md
andreabadesso: I agree this is confusing, refactored it to return in separate fields:
025194d18cb6293ae094d7dc542cb7ebe18d9689

projects/web-wallet/rpc-protocol.md
andreabadesso: I like the idea, done in 657cbc8a619b5de1cdeaa2ad1ae7ff9971ed0d20

andreabadesso: Yes, done in c9d9d7381b55c99bfb85700fc193af8026b2d2bf

The JSON-RPC implementation doesn't specify a hard limit for response sizes and I couldn't find any limits in the wallet connect documentation, but I limited it to 30 addresses at a time

What do you think about this new format?

projects/web-wallet/rpc-protocol.md
andreabadesso: I like this idea, also added a new type of error

Done in 7181885a8c9dc943a95ee1865c5ea31eea24ef01

projects/web-wallet/rpc-protocol.md
andreabadesso: What I mean here is that I'm suggesting that we don't allow the user to select which utxo he wants to return, he can only reject or approve

In the future, we might allow him to select exactly which one he approves and even replace it with another one

projects/web-wallet/rpc-protocol.md
andreabadesso: Done in f8c0b94cf09a39c35d648a0850ae201593ec0ec8

projects/web-wallet/rpc-protocol.md
andreabadesso: What do you mean? Modify the balance in the client before returning?

projects/web-wallet/rpc-protocol.md
andreabadesso: I don't think we should in this specific case, he already approved the dApp connection, fetching the network seems harmless to me

projects/web-wallet/rpc-protocol.md
msbrogli: Ok.

projects/web-wallet/rpc-protocol.md
msbrogli: Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: Got it. Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: Great! Thanks!

projects/web-wallet/rpc-protocol.md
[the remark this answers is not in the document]
msbrogli: Sounds great. Thanks!

projects/web-wallet/rpc-protocol.md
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

projects/web-wallet/rpc-protocol.md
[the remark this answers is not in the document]
msbrogli: It seems good to me. Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: It's fine. This new method will be called by the wallet, right?

projects/web-wallet/rpc-protocol.md
msbrogli: Thanks!

msbrogli: Can't the dApp request a ready-to-push transaction?

projects/web-wallet/rpc-protocol.md
msbrogli: Great. Thanks! I think we can add an example. For instance:

```
message = "This is my message"
signed_message = "Hathor Signed Message: This is my message"
```

projects/web-wallet/rpc-protocol.md
msbrogli: Great! Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: Great! Thanks!

msbrogli: It seems great! And I think the limit of 30 is good enough.

projects/web-wallet/rpc-protocol.md
msbrogli: Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: Got it. Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: Let's say I have 1,000 HTR in my wallet. I might be comfortable to share that I have a balance of 50 HTR (which is smaller than the actual balance).

projects/web-wallet/rpc-protocol.md
andreabadesso: Yes, the dApp will send a RPC to the wallet and the wallet will execute it

projects/web-wallet/rpc-protocol.md
andreabadesso: I like the idea, done in 0a62d03c35a04cd1c9a7540ccbf0954e91f40e23

projects/web-wallet/rpc-protocol.md
andreabadesso: Done in 2b4bdfe24dc2a67362e449fbee7dc0febb6be0af

What do you think about the format?

projects/web-wallet/rpc-protocol.md
andreabadesso: I like this idea, I added a new RPC call to fetch the status for an operation (using its id) and added a specific `op_status` to the push tx RPC call, which is the one that has multiple states

b8eb722066bfa1edacf674589251c835bbd264cc


What do you think about this format?

msbrogli: It seems great to me! Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: It seems great to me! Thanks!

projects/web-wallet/rpc-protocol.md
msbrogli: Maybe add that the chosen balance should never be higher than the current balance for a given token.

projects/web-wallet/rpc-protocol.md
msbrogli: Great! Thanks!

msbrogli (approved):

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,643 @@
...
+13. `network` - The network this dApp is expecting the client to be. The request
+   will fail if the client's network is different from this params value
+
+**Example Request:**
+
+```json
+{
+  "id": 3,
+  "jsonrpc": "2.0",
+  "method": "htr_signWithAddress",
```
pedroferreira1, line 173: Wrong method

projects/web-wallet/rpc-protocol.md
```diff
@@ -0,0 +1,643 @@
...
+```
+
+### 2.1.7 `htr_getAddress`
+
+This method allows the dApps to request an address.
+**Parameters**
+
+1. `type` string (Defaults to `client`) - Can be one of:
+
+`first_empty` should return the first empty address
```
pedroferreira1, line 412: The parse is bad for this syntax. Maybe we could have subitems.

image

projects/web-wallet/rpc-protocol.md
andreabadesso: Done 3f97ffc5d150fa0b9e22b0a16df80e5af2bd4378

projects/web-wallet/rpc-protocol.md
andreabadesso: Fixed in 14a7fa31876e1ca55bd63a3eb57fd26a98d43cd5

projects/web-wallet/rpc-protocol.md
andreabadesso: Fixed in eae856bcde08828c6aa8c5decb7cfd1bfa676e32

projects/web-wallet/rpc-protocol.md
```diff
@@ -641,3 +735,25 @@ An example from metamask we can take inspiration from:
...
+
+If the validation is successful, we will display a confirmation screen,
+with all outputs for manual user validation (UX is yet to be discussed). If the
+user rejects the transaction, we will respond with a `REQUEST_REJECTED` error
+
+The address will default to the address that is registered with the nano
+contract on the wallet or if it's not yet registered, it will default to the
+address with index 0.
+
+The user will be allowed to select any other address from his wallet or manually
```
pedroferreira1, line 754: This address is the address that will be used to sign the nano contract transaction, so the user can't manually input any valid address he wishes, he must choose one address from his wallet.

pedroferreira1 (dismissed):

projects/web-wallet/rpc-protocol.md
```diff
@@ -550,6 +550,100 @@ The `op_status` is a status specific to the operation being executed in this RPC
...
+
+If `push_tx` is enabled, it will select the `utxos` automatically, mine and send
+the transaction, otherwise, it will just return the transaction with signed
+inputs
+
+**Parameters**
+
+1. `method` - The nano contract method to call
+2. `blueprint_id` - The blueprint id of the nano contract
+3. `nc_id` - The nano contract instance id to call
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 95b99f751ed44a7f22e1ce376989b05a4ebb4cd4

projects/web-wallet/rpc-protocol.md
andreabadesso: Done in a5a5a07177b955da5df021a3d2e144368c61a434

pedroferreira1 (approved):

projects/web-wallet/rpc-protocol.md
msbrogli: Great! Thanks!

msbrogli (approved):

PR #56, opened 2023-06-12
Design: Automated Headless api-docs upload

tuliomir:
This feature automates the process of uploading the updated API Docs for the Headless Wallet, while giving the developer a preview of the contents, error checking and linting.

View [rendered](https://github.com/HathorNetwork/rfcs/blob/42b803d0ee690f90b9c19fe91b518b16f219a838/projects/hathor-wallet-headless/api-docs-ci.md)

### DIFF
--- a/projects/hathor-wallet-headless/api-docs-ci.md
+++ b/projects/hathor-wallet-headless/api-docs-ci.md
@@ -13,11 +13,12 @@ Currently this upload is made manually by a team member, extracting data from th
 The platform that manages the OpenAPI Documentation static website, [Redoc](https://github.com/Redocly/redoc), only interacts with `.json` files. So we need to convert our `src/api-docs.js` file into a `.json` one whenever we want to use this platform.
 
 A script should be built to make this conversion on the `scripts` folder, that implements the following:
-- Removes the `securitySchemes` from the `components` property
 - Ensures the `default` property normally exported by the `JSON.stringify()` is not present on the output json
 - Outputs the results to the `tmp/api-docs.json` folder.
 
 This script will be referred to as `scripts/convert-docs.js` from now on.
+
+> *Note*: The `securitySchemes` from the `components` property is no longer necessary and can be removed from the `js` file when this is implemented.
 
 ### Linting and Preview
 A developer should be able to run a `npm run docs` command with three outcomes:
--- a/projects/hathor-wallet-headless/api-docs-ci.md
+++ b/projects/hathor-wallet-headless/api-docs-ci.md
@@ -12,9 +12,10 @@ Currently this upload is made manually by a team member, extracting data from th
 ### JS to JSON conversion
 The platform that manages the OpenAPI Documentation static website, [Redoc](https://github.com/Redocly/redoc), only interacts with `.json` files. So we need to convert our `src/api-docs.js` file into a `.json` one whenever we want to use this platform.
 
-A script should be built to make this conversion on the `scripts` folder, outputting the results to the `tmp/` folder. All following steps will be executed on this converted file. Some specific conditions to execute:
-- Remove the `securitySchemes` from the `components` property
-- Ensure the `default` property normally exported by the javascript `.toJSON()` is not present on the output json
+A script should be built to make this conversion on the `scripts` folder, that implements the following:
+- Removes the `securitySchemes` from the `components` property
+- Ensures the `default` property normally exported by the `JSON.stringify()` is not present on the output json
+- Outputs the results to the `tmp/api-docs.json` folder.
 
 This script will be referred to as `scripts/convert-docs.js` from now on.
 
@@ -35,6 +36,9 @@ npx @redocly/cli lint tmp/api-docs.json
 # Local server for validating
 npx @redocly/cli preview-docs tmp/api-docs.json
 xdg-open http://127.0.0.1:8080
+
+# Cleanup after the server is closed
+rm tmp/api-docs.json
 ```
 
 > *Note:* The linter throws some errors on our current documentation. The first implementation of this script will certainly require a refactoring PR.
--- a/projects/hathor-wallet-headless/api-docs-ci.md
+++ b/projects/hathor-wallet-headless/api-docs-ci.md
@@ -48,9 +48,47 @@ rm tmp/api-docs.json
 A GitHub Workflow, configured to run on every PR, should run the conversion script and linter to check for errors on the generated documentation. This should be part of the CI process.
 
 ### Deployment
-On the deployment pipeline, namely when there is a new `v*` label on the `master` branch, a new workflow should also upload the generated `json` file to the production documentation S3 bucket. This will update the website to the community.
+On the deployment pipeline, namely when there is a new release label on the `master` branch, a new workflow should also upload the generated `json` file to the production documentation S3 bucket. This will update the website to the community.
+
+More on what consists a release label can be found on the reference-level explanation.
 
 > *Note:* All connection data with the S3 should be implemented as secrets. Refer to the [`docker.yml`](https://github.com/HathorNetwork/hathor-wallet-headless/blob/master/.github/workflows/docker.yml) file for reference.
+
+# Reference-level explanation
+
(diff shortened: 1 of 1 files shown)


projects/hathor-wallet-headless/api-docs-ci.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Motivation
+Currently this upload is made manually by a team member, extracting data from the `src/api-docs.js` file and processing it to finally upload its contents to the [Headless Hathor Wallet API](https://wallet-headless.docs.hathor.network/) site.
+
+# Guide-level explanation
+### JS to JSON conversion
+The platform that manages the OpenAPI Documentation static website, [Redoc](https://github.com/Redocly/redoc), only interacts with `.json` files. So we need to convert our `src/api-docs.js` file into a `.json` one whenever we want to use this platform.
+
+A script should be built to make this conversion on the `scripts` folder, outputting the results to the `tmp/` folder. All following steps will be executed on this converted file. Some specific conditions to execute:
+- Remove the `securitySchemes` from the `components` property
```
luislhl, line 16: Why do we need to remove it? And what are they used for if we will remove them? Shouldn't we remove them from the `src/api-docs.js` file?

projects/hathor-wallet-headless/api-docs-ci.md
```diff
@@ -0,0 +1,63 @@
...
+
+This script will be referred to as `scripts/convert-docs.js` from now on.
+
+### Linting and Preview
+A developer should be able to run a `npm run docs` command with three outcomes:
+- A validation check should be executed on the generated docs, alerting any errors found
+- A lint should be executed on the generated docs, displaying its analysis results
+- A local server should be fired, allowing the developer to visualize the results of the doc changes
+
+This would be an automated version of manually executing the following commands:
```
luislhl, line 27: Maybe we should make sure to remove the `tmp/api-docs.json` file at the end of the process.

projects/hathor-wallet-headless/api-docs-ci.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Local server for validating
+npx @redocly/cli preview-docs tmp/api-docs.json
+xdg-open http://127.0.0.1:8080
+```
+
+> *Note:* The linter throws some errors on our current documentation. The first implementation of this script will certainly require a refactoring PR.
+
+### PR Validation
+A GitHub Workflow, configured to run on every PR merging to `dev`, should run the conversion script and linter to check for errors on the generated documentation. This should be part of the CI process.
```
luislhl, line 43: Why only on PRs merging to `dev`? What if someone creates stacked PRs?

projects/hathor-wallet-headless/api-docs-ci.md
tuliomir: The current contents of this property are:
```json
"securitySchemes": {
  "ApiKeyAuth": {
    "type": "apiKey",
    "in": "header",
    "name": "X-API-KEY"
  }
}
```
So, it's not that we're sharing any sensitive information.
@pedroferreira1 , since you have more experience with this workflow, could you share with us why we currently remove this property before publishing the api-docs?

projects/hathor-wallet-headless/api-docs-ci.md
tuliomir: I see no negative consequence of having this file lingering at the `tmp` folder:
- On the local dev environment it could be reused by the `preview-docs` command
- On the CI environment, it would be destroyed as soon as the workflow is over

However I think it's a good practice to remove the tmp files if we have no explicit reason to keep it. :+1: 
Changed it on [9d4de6e](https://github.com/HathorNetwork/rfcs/pull/56/commits/9d4de6ebeee1a96b1e82876e2abbe7c127db52e9)

projects/hathor-wallet-headless/api-docs-ci.md
tuliomir: Our current default is to only run our workflows on PRs merging to `dev` ( or, in cases where this branch doesn't exist, on `master` / `main` ).
If it becomes really necessary to run this on a separate branch, it would always be possible to add the branch name on the workflow `yml` file on that specific branch / PR.

What do you think?

projects/hathor-wallet-headless/api-docs-ci.md
tuliomir: Removal suggested on [a336ca7](https://github.com/HathorNetwork/rfcs/pull/56/commits/a336ca7eefcaa335323fd15b6f1573613196db66)

projects/hathor-wallet-headless/api-docs-ci.md
luislhl: I think the problem with running tests/linters only for PRs based on `dev`/`master` is that PRs based on another feature branch wouldn't get their tests run.

This could lead us to introducing bugs when merging these PRs on top of one another, or at least taking longer to give some feedback for the developer about problems in their code.

For instance, I got one from our board to show you this is not so rare to happen: https://github.com/HathorNetwork/hathor-core/pull/645

This PR only had tests run on it because hathor-core is currently configured to run on any PR regardless of the base branch: https://github.com/HathorNetwork/hathor-core/blob/master/.github/workflows/main.yml#L10

tuliomir: While I think this would be less necessary for a documentation feature, I think the overhead is very small ( the validation proposed is very fast and light on resources ) and the benefit is clear on the CI.

Removed the branch restriction on [a6d072b](https://github.com/HathorNetwork/rfcs/pull/56/commits/a6d072beae15f7f4700f450c3b9d7daf2b0b6077)

luislhl (approved):

projects/hathor-wallet-headless/api-docs-ci.md
```diff
@@ -0,0 +1,68 @@
...
+```sh
+# Validation and conversion
+node ./scripts/convert-docs.js
+
+# Linting
+npx @redocly/cli lint tmp/api-docs.json
+
+# Local server for validating
+npx @redocly/cli preview-docs tmp/api-docs.json
+xdg-open http://127.0.0.1:8080
```
pedroferreira1, line 39: Why do you need this? The redoc-cli can generate an html file, then you can just `open index.html` and the developer can see the generated file in the browser.

Better than adding another dependency in my opinion, even if it's a dev dependency

projects/hathor-wallet-headless/api-docs-ci.md
```diff
@@ -0,0 +1,68 @@
...
+rm tmp/api-docs.json
+```
+
+> *Note:* The linter throws some errors on our current documentation. The first implementation of this script will certainly require a refactoring PR.
+
+### PR Validation
+A GitHub Workflow, configured to run on every PR, should run the conversion script and linter to check for errors on the generated documentation. This should be part of the CI process.
+
+### Deployment
+On the deployment pipeline, namely when there is a new `v*` label on the `master` branch, a new workflow should also upload the generated `json` file to the production documentation S3 bucket. This will update the website to the community.
```
pedroferreira1, line 51: rc tags shouldn't trigger this workflow.

projects/hathor-wallet-headless/api-docs-ci.md
tuliomir: `xdg-open` is the Ubuntu version for Mac's `open` command. So, no dependency will be added: it's only a request to open the browser on the local server that the cli starts.

The only dependency that will be added here is, indeed, the redoc-cli ( named [`@redocly/cli`](https://redocly.com/docs/cli/) )

projects/hathor-wallet-headless/api-docs-ci.md
tuliomir: Good point. Added further explanation to clarify this on [42b803d](https://github.com/HathorNetwork/rfcs/pull/56/commits/42b803d0ee690f90b9c19fe91b518b16f219a838)

pedroferreira1 (approved):

PR #57, opened 2023-06-16
design: extend MultiSig wallet

alexruzenhack:
Extend the MultiSig wallet to include the following features:
- create token
- mint token
- melt token
- extend decode transaction
- start read-only wallet

[Rendered](https://github.com/HathorNetwork/rfcs/blob/36e91ad4a47f6364043b3a5b430400fde971de56/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md)

Closes https://github.com/HathorNetwork/hathor-wallet-headless/issues/296

### DIFF
--- a/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
+++ b/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
@@ -41,10 +41,10 @@ The features we need to add for the MultiSig wallet:
 
 Taking by example the `POST:/wallet/p2sh/tx-proposal` and `POST:/wallet/p2sh/tx-proposal/get-my-signatures` endpoints we can identify the following semantic:
 
-`/wallet` -- the wallet module
-`/p2sh` -- a submodule representing a domain of wallet
-`/tx-proposal` -- a command to create a new transaction proposal
-`/tx-proposal/get-my-signatures` -- this time `/tx-proposal` assumes the role of a component of p2sh domain in the wallet module, and `/get-my-signatures` is the command to sign the transaction with the wallet private key
+- `/wallet` -- the wallet module
+- `/p2sh` -- a submodule representing a domain of wallet
+- `/tx-proposal` -- a command to create a new transaction proposal
+- `/tx-proposal/get-my-signatures` -- this time `/tx-proposal` assumes the role of a component of p2sh domain in the wallet module, and `/get-my-signatures` is the command to sign the transaction with the wallet private key
 
 In summary, we can have either syntax:
 
@@ -159,7 +159,7 @@ The response scheme for success:
 ```ts
 {
   "success": boolean,
-  "complete": boolean,
+  "completeSignatures": boolean,
   "tx": {
     "version": number,
     "tokens": string[],
@@ -210,7 +210,7 @@ The response scheme for success:
 ```
 
 Lets represent the response document as ` $ `, and each element of a list ` [*] `, we have:
-* `$.complete` -- that represents the completeness of signatures required to use the transaction
+* `$.completeSignatures` -- that represents the completeness of signatures required to use the transaction
 * `$.tx.inputs[*].signed` -- that indicates this input has a signature
 * `$.tx.inputs[*].mine` -- that indicates this input belongs to this wallet
 * `$.tx.outputs[*].decoded.mine` -- that indicates this output belongs to this wallet
--- a/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
+++ b/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
@@ -34,7 +34,7 @@ The features we need to add for the MultiSig wallet:
 * Create a custom token
 * Mint a custom token
 * Melt a custom token
-* Decode transaction with wallet metadata
+* Inspect transaction with wallet metadata
 * Initialize wallet without seed
 
 ## API endpoints design
@@ -97,8 +97,7 @@ For the MultiSig wallet the endpoint keeps the same command:
 The data fields are:
 * `token`*
 * `amount`*
-* `address`*
-* `create_mint`
+* `address`
 * `change_address`
 * `mint_authority_address`
 * `allow_external_mint_authority_address`
@@ -128,7 +127,6 @@ The data fields are:
 * `token`*
 * `amount`*
 * `deposit_address`
-* `create_melt`
 * `change_address`
 * `melt_authority_address`
 * `allow_external_melt_authority_address`
@@ -145,7 +143,7 @@ The response scheme for success:
 
 The purpose of the command is only produce the `txHex` not signed and nothing more, it will not push the transaction to the network. The returned `txHex` will be handled by the coordinator service, and later on by the federation service, which is responsible to sign the transaction.
 
-With this specification we cover the possibility to melt tokens, send it to another wallet, create or not another melt authority, and send the authority to another wallet.
+With this specification we cover the possibility to melt tokens, send it to another wallet, create or not another melt authority, send the authority to another wallet, and define an address to deposit the withdraw amount if any.
 
 ## Decode transaction with wallet metadata
 
@@ -161,54 +159,61 @@ The response scheme for success:
 ```ts
 {
   "success": boolean,
-  "completeSignatures": boolean,
+  "complete": boolean,
   "tx": {
     "version": number,
     "tokens": string[],
     "inputs": [
       {
-        tx_id: string,
+	    txId: string,
         index: number,
-        decoded: {},  // data from the spent ouput
+        decoded: {
+            type: string,
+            address: string,
+            timelock: number,
+        },  // data from the spent ouput
+        token: string,
+        value: number,
+        tokenData: number, // user face
+        token_data: number, // internal use
+        script: string,
         signed: boolean,
+        mine: boolean,
       },
     ],
     "outputs": [
       {
+	    decoded: {
+		    address: string,
+		    timelock: number,
+		    mine: boolean,
+	    },
         value: number,
-        token_data: number,
+        tokenData: number, // user face
+        token_data: number, // internal use
         script: string,
-        decoded: {},
-        authorities: number,
-        mine: boolean,
+        type: string,
+        token?: string,
       }
     ],
   },
   "balance": {
     "<token-uid>":  {
-      "tokens": {
-        available: number,
-        locked: number
-       },
+      "tokens": { available: number, locked: number ),
       "authorities": {
-        mint: {
-          available: number,
-          locked: number
-        },
-        melt: {
-          available: number,
-          locked: number
-        },
+        mint: { available: number, locked: number ),
+        melt: { available: number, locked: number ),
       },
     },
   },
 }
 ```
 
-Lets represent the response document as ` $ `, as each element of a list ` [*] `, we have:
-* `$.completeSignatures` -- that represents the completeness of signatures required to use the transaction
+Lets represent the response document as ` $ `, and each element of a list ` [*] `, we have:
+* `$.complete` -- that represents the completeness of signatures required to use the transaction
 * `$.tx.inputs[*].signed` -- that indicates this input has a signature
-* `$.tx.outputs[*].mine` -- that indicates this output belongs to this wallet
+* `$.tx.inputs[*].mine` -- that indicates this input belongs to this wallet
+* `$.tx.outputs[*].decoded.mine` -- that indicates this output belongs to this wallet
 * `$.balance[*].tokens` -- that contains the balance for the token
 * `$.balance[*].authorities` -- that contains the balance of mint and melt for the token
 
--- a/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
+++ b/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
@@ -34,7 +34,7 @@ The features we need to add for the MultiSig wallet:
 * Create a custom token
 * Mint a custom token
 * Melt a custom token
-* Inspect transaction with wallet metadata
+* Decode transaction with wallet metadata
 * Initialize wallet without seed
 
 ## API endpoints design
@@ -98,6 +98,7 @@ The data fields are:
 * `token`*
 * `amount`*
 * `address`*
+* `create_mint`
 * `change_address`
 * `mint_authority_address`
 * `allow_external_mint_authority_address`
@@ -126,7 +127,8 @@ For the MultiSig wallet the endpoint keeps the same command:
 The data fields are:
 * `token`*
 * `amount`*
-* `address`*
+* `deposit_address`
+* `create_melt`
 * `change_address`
 * `melt_authority_address`
 * `allow_external_melt_authority_address`
@@ -159,7 +161,7 @@ The response scheme for success:
 ```ts
 {
   "success": boolean,
-  "complete": boolean,
+  "completeSignatures": boolean,
   "tx": {
     "version": number,
     "tokens": string[],
@@ -169,7 +171,6 @@ The response scheme for success:
         index: number,
         decoded: {},  // data from the spent ouput
         signed: boolean,
-        mine: boolean,
       },
     ],
     "outputs": [
@@ -185,10 +186,19 @@ The response scheme for success:
   },
   "balance": {
     "<token-uid>":  {
-      "tokens": { available: number, locked: number ),
+      "tokens": {
+        available: number,
+        locked: number
+       },
       "authorities": {
-        mint: { available: number, locked: number ),
-        melt: { available: number, locked: number ),
+        mint: {
+          available: number,
+          locked: number
+        },
+        melt: {
+          available: number,
+          locked: number
+        },
       },
     },
   },
@@ -196,9 +206,8 @@ The response scheme for success:
 ```
 
 Lets represent the response document as ` $ `, as each element of a list ` [*] `, we have:
-* `$.complete` -- that represents the completeness of signatures required to use the transaction
+* `$.completeSignatures` -- that represents the completeness of signatures required to use the transaction
 * `$.tx.inputs[*].signed` -- that indicates this input has a signature
-* `$.tx.inputs[*].mine` -- that indicates this input belongs to this wallet
 * `$.tx.outputs[*].mine` -- that indicates this output belongs to this wallet
 * `$.balance[*].tokens` -- that contains the balance for the token
 * `$.balance[*].authorities` -- that contains the balance of mint and melt for the token
--- a/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
+++ b/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
@@ -1,0 +1,310 @@
+- Feature Name: MultiSig to assist the EVM compatible bridge by supporting create, mint and melt token commands
+- Start Date: 2023-06-14
+- RFC PR: (leave this empty)
+- Hathor Issue: (leave this empty)
+- Author: Alex Ruzenhack alex@hathor.network
+
+#### Table of content
+[table-of-content]: #table-of-content
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Current system](#current-system)
+- [Guide-level explanation](#guide-level-explanation)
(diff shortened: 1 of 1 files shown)


projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
```diff
@@ -0,0 +1,327 @@
...
+The purpose of the command is only produce the `txHex` not signed and nothing more, it will not push the transaction to the network. The returned `txHex` will be handled by the coordinator service, and later on by the federation service, which is responsible to sign the transaction.
+
+With this specification we cover the possibility to melt tokens, send it to another wallet, create or not another melt authority, and send the authority to another wallet.
+
+## Inspect transaction with wallet metadata
+
+There is an endpoint to [inspect the transaction](https://hathor.gitbook.io/hathor/guides/multisig-wallet/sending-transactions#5.-check-the-multisig-transaction) by decoding the transaction hexadecimal encode in `POST:/wallet/decode`. However, it doesn't bring signature information, nor wallet information with it.
+
+For the MultiSig wallet we have the `inspect` command:
+`POST:/wallet/p2sh/tx-proposal/inspect`
```
pedroferreira1, line 153: Is there any reason to create a new endpoint, instead of just using the same (`/wallet/decode`) expanding it to return a more detailed data? Maybe even with a parameter `include_metadata` if we feel the common use is to get only the simple data.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
```diff
@@ -0,0 +1,327 @@
...
+* `$.balance.tokens` -- that contains the balance of each token in the transaction
+* `$.balance.authorities` -- that contains the balance of mint and melt for each token
+
+## Initialize MultiSig wallet without seed
+
+This wallet is used by the coordinator service to sign or sign and push the transaction given the signatures collected by the service.
+
+By relaxing the checks in `/wallet/start` endpoint logic a wallet can have in its configuration the `multisig` key without the `seeds` key. This configuration impairs the wallet to use `/wallet/p2sh/tx-proposal/get-my-signatures` endpoint, however it still allows the use of `/wallet/p2sh/tx-proposal/sign` and `/wallet/p2sh/tx-proposal/sign-and-push` endpoints because the signatures are provided in the call.
+
+This wallet can also be called the "seedless wallet".
```
pedroferreira1, line 212: I would like to understand a bit the low level of this feature. The sign endpoint calls a method that validates the utxos of the multisig wallet, for instance, so you would need to create a readonly wallet with balance, history and utxos loaded, not only a seedless wallet. That's why I feel a better description of the low level would be good. The lib already supports a readonly wallet for multisig?

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
```diff
@@ -0,0 +1,327 @@
...
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+* Create authority for a given token
+* Destroy authority for a given token
+* Delegate authority for a given token
+* Create NFT token
+
+## Create authority for a given token
```
pedroferreira1, line 243: I didn't understand this feature, can you elaborate more?

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
r4mmer: I don't think the syntax consistency is broken when we use the `/wallet/decode` api, we will improve an api made to work with both p2pkh and p2sh wallets.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Some exchange or other partner may create authority to mint or melt and spread it though its wallets for the purpose to increase throughput or spread governance power.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
r4mmer: The idea is that we only require the multisig configuration to derive the addresses, so we can know the balance of the wallet and receive transactions in real time.
If we receive the signatures from the actual participants and we have the multisig configuration we can derive the redeem script and input data, making the transaction ready to be mined.

But I agree the design should include how we are going to do this.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
```diff
@@ -0,0 +1,327 @@
...
+}
+```
+
+Lets represent the response document as ` $ `, as each element of a list ` [*] `, we have:
+* `$.complete` -- that represents the completeness of signatures required to use the transaction
+* `$.tx.inputs[*].signed` -- that indicates this input has a signature
+* `$.tx.inputs[*].mine` -- that indicates this input belongs to this wallet
+* `$.tx.outputs[*].mine` -- that indicates this output belongs to this wallet
+* `$.balance.tokens` -- that contains the balance of each token in the transaction
+* `$.balance.authorities` -- that contains the balance of mint and melt for each token
```
r4mmer, line 204: `tokens` and `authorities` are for a single token, and the `balance` is a map where the keys are all tokens in the transaction.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
pedroferreira1: The reason for my comment is that the tx decode doesn't change if we are using a multisig wallet or not. We have different endpoints for sending transactions because there are some specific steps when sending a tx with a multisig wallet but decoding a transaction is always the same.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
pedroferreira1: The feature described is a delegate authority, which is described in the other section.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Well, if the API consistency is not a thing in the way I have though, and we are ok to change the scheme for the response in the presence of "include_metadata" param in the request, we can extend the `/wallet/decode`.

Particularly I consider it a breaking change, and it would be difficult to describe and explain this subtle change in the scheme with the introduction of a request param.

alexruzenhack: My proposition to solve this topic is:
- extend the `/wallet/decode` to multisig -- if needed
- create a second endpoint `/wallet/enriched-decode` that contains wallet metadata

For future should we think to have a GraphQL like endpoint?

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Oh, no. In the delegate authority we "transfer" the ownership of an authority without "create" new ones, right? In the create authority we are exchanging 1 authority to N authorities. They are not the same thing, except if it is possible to "create" N authorities with the delegate as well. If it is the case, you right.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Working on it.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
r4mmer: I like having a second endpoint for this, but I think `/wallet/inspect` seems more to the point. What do you think?

pedroferreira1: The information added as metadata seems to be useful even for normal wallets and use. I prefer that we expand our response object to add this new data. With that we don't create a new endpoint specifically for metadata (this would require users to send two requests to get full data), and we don't add breaking changes, because the new data will be a complement of the current response data.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
pedroferreira1: In the end, we always create new authorities because we must spend one authority utxo. By default we create 2 authorities, one new and one for yourself, but we can create 1 or 2 with the current lib code. I understand that the proposal is to be able to create more than 2, and I like the idea, I just think it's the same as the delegate authority, only more powerful.

alexruzenhack: Ok. Vou descrever então como uma extensão do delegate.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: I will fix it.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: You are not considering this addition as a breaking change because nothing would change for the calls without the `include_metadata` param, right?

I consider any change in the response scheme a breaking change, even complementing ones, because one may validate the response scheme before use it.

I don't see any problem for us to proceed with the `include_metadata` solution. But we should agree that any change in the `include_metadata` scheme in the future would be a breaking change, except if we add another mechanism to not disrupt the current scheme usage, and to avoid these exceptional mechanisms I believe we need a versioning system for the API.

But coming back to the problem of documentation to include two schemes for the same endpoint, the OpenAPI v3 supports the ["Discriminator Object"](https://swagger.io/specification/#discriminator-object), which is the specification for multiple schemes, however it doesn't support deliver an specific scheme related to a param and we should write a very clear documentation about this scheme change attached to the `include_metadata`.

---

Facing these limitations of OpenAPI, the extra care with the documentation to bring clarity and the creation of exceptional mechanism to keep backward compatibility, I'm starting to gain resistance toward this solution.

The creation of `/wallet/inspect` endpoint seen now pretty neat, more simple. We can still have only one function to decode.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Fixed

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Done.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: I found that we don't need change anything. I updated the topic "Initialize read-only MultiSig wallet".

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: I have [updated the design](https://github.com/HathorNetwork/rfcs/blob/24ec11b13901870c115774bb117ba962f8079851/projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md#decode-transaction-with-wallet-metadata) with the decision we made together.

alexruzenhack:
@pedroferreira1 @r4mmer 

I realize that methods to build the multisig transactions in the lib generate signatures and set it in the input, therefore the txHex comes with `input.data`. However, despite the txHex comes with `input.data` all the transactions worked just fine. Even still, I decided to remove the `input.data` to not confuse the user when one use the `/wallet/decode`. Without removing the `input.data`, the decode would inform the txHex as being complete.

See: https://github.com/HathorNetwork/hathor-wallet-headless/pull/309/commits/c2f37645049d7cb4633b4a7732a167d31fd600ed#diff-a5cc140e5cdfce35607f3618c60a83b9b7f1e5fa5c0e84bd30b49c3717e3ad5eR91-R94

However, I think this is a quick solution to not over extend my time here at this project. The best solution would be implement build methods in the lib that do not sign the transaction, in order to return a pure transaction proposal.

r4mmer:
> @pedroferreira1 @r4mmer
> 
> I realize that methods to build the multisig transactions in the lib generate signatures and set it in the input, therefore the txHex comes with `input.data`. However, despite the txHex comes with `input.data` all the transactions worked just fine. Even still, I decided to remove the `input.data` to not confuse the user when one use the `/wallet/decode`. Without removing the `input.data`, the decode would inform the txHex as being complete.
> 
> See: [HathorNetwork/hathor-wallet-headless@c2f3764#diff-a5cc140e5cdfce35607f3618c60a83b9b7f1e5fa5c0e84bd30b49c3717e3ad5eR91-R94](https://github.com/HathorNetwork/hathor-wallet-headless/commit/c2f37645049d7cb4633b4a7732a167d31fd600ed#diff-a5cc140e5cdfce35607f3618c60a83b9b7f1e5fa5c0e84bd30b49c3717e3ad5eR91-R94)
> 
> However, I think this is a quick solution to not over extend my time here at this project. The best solution would be implement build methods in the lib that do not sign the transaction, in order to return a pure transaction proposal. I believe it should happen in another project with a proper design to alter the wallet-lib.

I think we have 2 options, we can check if the wallet is multisig then don't sign in all methods or make the signing part optional with a parameter in all methods.
I like the second option more, if the option is true by default it would not change the current behavior

alexruzenhack:
> I think we have 2 options, we can check if the wallet is multisig then don't sign in all methods or make the signing part optional with a parameter in all methods.
> I like the second option more, if the option is true by default it would not change the current behavior

I like this second option as well. If we decide to follow this path, the change will be minimal and I can implement it in this project.

alexruzenhack:
> I like this second option as well. If we decide to follow this path, the change will be minimal and I can implement it in this project.

I implemented the second option in the wallet-lib.

pedroferreira1:
> I like this second option as well. If we decide to follow this path, the change will be minimal and I can implement it in this project.

I agree

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
```diff
@@ -0,0 +1,319 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+* This design is a complement for the  [EVM compatible bridge design](https://github.com/HathorNetwork/rfcs/blob/doc/evm-compatible-brigde/projects/evm-compatible-bridge/design.md#federation-service). 
+* The commands to create token, mint and melt includes a fine tuned authority creation after the PRs [#291](https://github.com/HathorNetwork/hathor-wallet-headless/pull/291) and [#293](https://github.com/HathorNetwork/hathor-wallet-headless/pull/293) in the wallet-headless.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+* Should NFT data be allowed in the create token endpoint for the MultiSig wallet, or do it deserves its own endpoint, as it happens for the single wallet?
```
tuliomir, line 253: How would that be helpful? Is there a use case in sight for this?

alexruzenhack: Hmm... The use case I have in sight is this: an organization that needs to mint a new NFT collection but use a multisig wallet. I far as I know, we don't have a partner that need it, but this is a possibility.

tuliomir (approved):

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
```diff
@@ -0,0 +1,324 @@
...
+`/tx-proposal` -- a command to create a new transaction proposal
+`/tx-proposal/get-my-signatures` -- this time `/tx-proposal` assumes the role of a component of p2sh domain in the wallet module, and `/get-my-signatures` is the command to sign the transaction with the wallet private key
+
+In summary, we can have either syntax:
+
+* `/module/domain/command`, or
+* `/module/domain/component/command`
+
+As our intend is to execute commands, the verb `POST` is the most convenient method to call the endpoints.
+
```
obiyankenobi, line 55: I'm wondering about the endpoints and if we should reorganize them. Doesn't seem intuitive that we have
`POST /wallet/p2sh/tx-proposal/get-my-signatures`
`POST /wallet/p2sh/tx-proposal/create-token`
`POST /wallet/p2sh/tx-proposal/mint-token`
`POST /wallet/p2sh/tx-proposal/melt-token`

But we use `POST /wallet/p2sh/tx-proposal/` for sending tokens. It seems more intuitive that we have `POST /wallet/p2sh/tx-proposal/send-token`.

We could add this API and continue supporting the current one, but mark as deprecated.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
```diff
@@ -0,0 +1,324 @@
...
+        mint: { available: number, locked: number ),
+        melt: { available: number, locked: number ),
+      },
+    },
+  },
+}
+```
+
+Lets represent the response document as ` $ `, and each element of a list ` [*] `, we have:
+* `$.complete` -- that represents the completeness of signatures required to use the transaction
```
obiyankenobi, line 213: Should we go with something that makes more sense? Like `has_all_signatures`. Or do we use this term (`complete`) for this same context already in other places?

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
```diff
@@ -0,0 +1,324 @@
...
+* `$.tx.inputs[*].signed` -- that indicates this input has a signature
+* `$.tx.inputs[*].mine` -- that indicates this input belongs to this wallet
+* `$.tx.outputs[*].decoded.mine` -- that indicates this output belongs to this wallet
```
obiyankenobi, line 216: Why is it inside the `decoded` section here but not on the input?

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
obiyankenobi: This is the same as checking all `$.tx.inputs[*].signed == true`, right?

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Ok! As we already have everything implemented, I can map as a "future possibility" and open an issue to make this change. What do you think?

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Oh! I forgot to update it here, now we call this attribute `completeSignatures`.

projects/hathor-wallet-headless/0001-multisig-create-mint-melt-token.md
alexruzenhack: Well, it happens because the output can be of type `data` and its structure doesn't have `address`. This way we would have the `mine` attribute with values `undefined` or `null`.
```
export interface IDataOutputData {
  type: 'data';
  token: string;
  value: number;
  authorities: number;
  data: string;
};
```

alexruzenhack: If you see it should be in the body as it happens with the input we can do it. I can add an Appendix to the design, open an issue and implement in another PR. What do you think?

I have opened a bump PR for the wallet-headless version already, but we can do it before the bump. Here is the bump PR: https://github.com/HathorNetwork/hathor-wallet-mobile/pull/304

obiyankenobi: Even when the output is of type data, shouldn't `mine` be `false` as well?

r4mmer: This makes sense, we should have the mine on `output.mine` and not on `output.decoded.mine` and if the output is a data output it will be false since a data output does not "belong" to any address.

obiyankenobi (approved):

PR #59, opened 2023-07-03
feat(feature-activation): implement bit signaling RFC

glevco:
This RFC describes a way for miners to signal feature support through the full node CLI. A POC is available in https://github.com/HathorNetwork/hathor-core/pull/688


projects/feature-activation/0002-bit-signaling.md
```diff
@@ -0,0 +1,64 @@
...
+[Guide-level explanation]: #guide-level-explanation
+
+Users running the full node, or specifically miners in this case, should be able to pass two new options to the `run_node` CLI interface:
+
+1. `--signal-support [feature]`, and
+2. `--signal-not-support [feature]`
+
+Here are rule definitions for these options:
+
+1. Both options receive the name of some feature, as defined in the existing `Feature` enum;
```
msbrogli, line 27: Should we have a `--list-signal-features`?

projects/feature-activation/0002-bit-signaling.md
```diff
@@ -0,0 +1,64 @@
...
+2. Both options can be used multiple times to either enable or disable support for multiple features;
```
msbrogli, line 28: How does it play with env vars? Maybe add a few examples to clarify.

projects/feature-activation/0002-bit-signaling.md
```diff
@@ -0,0 +1,64 @@
...
+3. A feature cannot be set in both options at the same time, as this would be a conflicting configuration;
+4. A feature can only be set in any of these options if it's currently undergoing a signaling phase of its Feature Activation process, otherwise setting its bit could conflict with other features;
```
msbrogli, line 30: I don't get it. The full node might warning if a feature has already been evaluated but I don't think the full node should fail its initialization if the signaling phase hasn't started or has already finished.

projects/feature-activation/0002-bit-signaling.md
```diff
@@ -0,0 +1,64 @@
...
+5. A default bit signal value (either enabled or disabled) should be provided for each feature.
```
msbrogli, line 31: Great! I'd say the full node should log the situation for each feature, which might be one of the following: `{ supported, not supported, supported (default), not supported (default) }`.

projects/feature-activation/0002-bit-signaling.md
```diff
@@ -0,0 +1,64 @@
...
+
+Here's an example of running the full node with those options, enabling support for features 2 and 3, and disabling support for feature 1:
+
+```bash
+poetry run hathor-cli run_node --status 8080 --testnet --memory-storage --signal-not-support NOP_FEATURE_1 --signal-support NOP_FEATURE_2 --signal-support NOP_FEATURE_3
+```
+
+Let's assume all features are disabled by default, and that `NOP_FEATURE_1` is configured to use `bit=0`, `NOP_FEATURE_2` is configured to use `bit=1`, and `NOP_FEATURE_3` is configured to use `bit=2`, where `bit=0` is the LSB, as explained in the previous RFC.
+
+Then, the signal bits for this full node run would be `110`, meaning that those bits would be present in all block templates generated for mining.
```
msbrogli, line 41: Only for the signaling phase of each of them, right? When out of signaling phase, the bits would be available for other features or they would simply be zero.

projects/feature-activation/0002-bit-signaling.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Reference-level explanation
+[Reference-level explanation]: #reference-level-explanation
+
+## Bit Signaling Serivce
```
msbrogli, line 46: typo

projects/feature-activation/0002-bit-signaling.md
```diff
@@ -0,0 +1,64 @@
...
+
+That class should also perform validations on the configuration, as specified in the list in the [Guide-level explanation] section. The full node must not start if those validations fail.
+
+## New signal_support_by_default configuration
+
+A new attribute called `signal_support_by_default` should be implemented in the `Criteria` class. It is a boolean that represents whether a feature should be supported by default or not, according to its configuration via `HathorSettings`. Its default value should be `False`.
+
+That is, assuming `signal_support_by_default=True` for `NOP_FEATURE_1`, and no CLI option is provided for that feature, then that feature is going to be signaled as supported in block templates. If some CLI option is set for that feature, then it takes precedence over the `signal_support_by_default` attribute.
+
+## Hathor Manager and mining APIs
```
msbrogli, line 58: All block template generators must also be updated accordingly.

msbrogli (changes requested):

projects/feature-activation/0002-bit-signaling.md
glevco: The list of features will appear when we run the `run_node --help` command, and also if the user sets a wrong string, the CLI helper will catch that and inform the user. I think that's enough and better than having a dedicated `--list-signal-features` command, what do you think?

projects/feature-activation/0002-bit-signaling.md
glevco: As implemented in the POC, the signal bits would be generated only once, during the full node initialization, and those same bits would be emitted for the whole duration of the full node run, even if the state of a feature changes.

The reason for this is that if we were to query the feature states and change the bits accordingly for each mined block, we would incur some new calculations in the mining API that could delay it, and I'm not sure that's desirable. I would rather leave it untouched, specially before we perform feature activation tests in testnet.

In other words, in the current code everything that is related to feature activation is completely independent of the full node code, meaning there's no current full node code that calls any feature activation method. The code for this RFC will be the first time some existing method in the full node calls some feature activation method. I believe there are two options:

1. Like in the POC, the only time feature activation states are calculated is during the full node initialization. So if there's any performance hit, it shouldn't be a problem. Then, the bits are stored and the mining block requests simply use them.
2. Or, we could calculate signal bits dynamically for each mining block request, that is, calculate feature states for each request. That would be required to stop enabling a bit when its feature becomes `ACTIVE`. That would introduce calculations that would happen during mining.

The tradeoff here is that by going with option 1, there's a possibility that some miner signal would "spill over" to another feature that reuses the same bit. Let's consider this possibility. We currently have 4 signal bits available, meaning that we can have 4 different feature activation processes running in parallel, without conflicting bits. That number is also configurable up to 8 bits, if we need more parallel features. On top of that, we also enforce at least one evaluation interval (2 weeks) before reusing a bit is allowed.

So a spill over could only happen if we ever have 4 parallel feature activation processes, and have a fifth feature reusing a bit 2 weeks later, and for that whole period the miner doesn't restart the full node. In practice, we should try to leave even more time between features before reusing a bit.

Considering the performance point, I believe the tradeoff is worth it. If we didn't provide this CLI convenience, the miners would have to set the bits themselves. In that case, the same spill over could happen before miners manually changed the bits they're setting. Since this tool is just a convenience, I think it would be better if it didn't affect the mining API. What do you think?

projects/feature-activation/0002-bit-signaling.md
glevco: I've added the example equivalent using env vars in https://github.com/HathorNetwork/rfcs/pull/59/commits/c0ec7e86771268e26cb2adf63bc51a31942fd3f1

projects/feature-activation/0002-bit-signaling.md
glevco: Sure, changed in https://github.com/HathorNetwork/rfcs/pull/59/commits/c0ec7e86771268e26cb2adf63bc51a31942fd3f1

projects/feature-activation/0002-bit-signaling.md
glevco: I've added a note to cover that in https://github.com/HathorNetwork/rfcs/pull/59/commits/c0ec7e86771268e26cb2adf63bc51a31942fd3f1

projects/feature-activation/0002-bit-signaling.md
glevco: Thanks! Fixed in https://github.com/HathorNetwork/rfcs/pull/59/commits/c0ec7e86771268e26cb2adf63bc51a31942fd3f1

projects/feature-activation/0002-bit-signaling.md
glevco: I added that line in https://github.com/HathorNetwork/rfcs/pull/59/commits/c0ec7e86771268e26cb2adf63bc51a31942fd3f1, but during implementation I'll have to reach out to you and/or Jan for specifics, to make sure we cover all template APIs

projects/feature-activation/0002-bit-signaling.md
msbrogli: I agree. I'm wondering where we can add this piece of information. But I agree it's a good solution.

projects/feature-activation/0002-bit-signaling.md
msbrogli: Thanks!

projects/feature-activation/0002-bit-signaling.md
msbrogli: Ok.

projects/feature-activation/0002-bit-signaling.md
msbrogli: Ok. We can do it at design time as well. I think every mining API uses the same block template generator. But we have to confirm it.

projects/feature-activation/0002-bit-signaling.md
msbrogli: Great! Thanks!

projects/feature-activation/0002-bit-signaling.md
glevco: Maybe in our GitBook documentation would be enough?

projects/feature-activation/0002-bit-signaling.md
glevco: Sure! In the POC I changed the `BlockTemplate(NamedTuple)` to add a new `signal_bits` attribute. This reflected in a change in `HathorManager`, that calls the new `BitSignalingService.get_signal_bits()` to set them in the `BlockTemplate`. But I'll check if that covers all mining APIs.

projects/feature-activation/0002-bit-signaling.md
msbrogli: As discussed in person, this behavior will be changed.

glevco: Indeed, I have updated the text to reflect our discussion in https://github.com/HathorNetwork/rfcs/pull/59/commits/674770f6c47e8ed42a3bea70d72e4d9c4fd84fd3

projects/feature-activation/0002-bit-signaling.md
glevco: I checked it, and doing it like it's done in the POC is enough to cover all mining APIs. Updated the text in https://github.com/HathorNetwork/rfcs/pull/59/commits/674770f6c47e8ed42a3bea70d72e4d9c4fd84fd3

projects/feature-activation/0002-bit-signaling.md
msbrogli: Great! Thanks!

projects/feature-activation/0002-bit-signaling.md
msbrogli: Good job! Thanks!

msbrogli (approved):

jansegre (approved):

PR #61, opened 2023-07-31
feat: using reliable integrations on the wallet-service

andreabadesso:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/feat/wallet-service-reliable-integration/projects/wallet-service-reliable-integration/0001-design.md)

### DIFF
--- a/projects/wallet-service-reliable-integration/0001-design.md
+++ b/projects/wallet-service-reliable-integration/0001-design.md
@@ -249,3 +249,34 @@ If we do have the transaction on the database, we need to check both the `voided
 
 - [ ] Maybe send the outputs already decoded, just like in the transactions API
 
+
+# Future Considerations
+
+## Simplified balance calculation
+
+There is a potential optimization to handling events by using the `group_id` attribute in our advantage for processing events in parallel. The approach can be summarized as follows:
+
+1. Group events by group_id.
+1. For each group of events, start a new database transaction by executing BEGIN TRANSACTION.
+1. Process each event individually. When an output is unspent, add an entry to the UTXO table. If an output is spent, remove the corresponding UTXO.
+  1. Note: Inputs or parents related actions are not necessary in this step
+  1. Events can be processed in parallel if and only if they're from different transactions.
+1. After processing all events in a group:
+  1. Generate new addresses.
+  1. Unlock any time-locked UTXOs.
+  1. Recalculate the balance.
+  1. Commit the changes to the database using COMMIT.
+
+We decided not to implement it in the initial version because we don't yet have a mechanism to know when a group has started or ended.
+
+## Reduced database reads with hashing
+
+Another potential enhancement to improve performance and reduce unnecessary database queries revolves around the use of hashing techniques, something similar to how "ETAG" used by modern browsers for detecting when a resource has changed.
+
+The proposal is to generate a hash, using a swift hashing algorithm like md5 or any other that might be faster yet equally efficient, on specific attributes we're keen on comparing (e.g., voided_by, height).
+
+By storing and comparing these hashes, it becomes feasible to quickly determine if there have been changes in the pertinent attributes without having to perform a full query or comparison.
+
+The comparison should be done by using a in-memory Least Recently Used (LRU) cache of these "e-tags" indexed by the transaction id. This cache should store up to 100,000 transaction hashes (configurable), ensuring quick access to the most frequently used or recently accessed hashes. 
+
+This approach can significantly decrease the load on the database, especially in situations where frequent checks or comparisons are required, but actual changes to the data of interest are infrequent (as is the case with multiple transactions being added as children for another transaction)
--- a/projects/wallet-service-reliable-integration/0001-design.md
+++ b/projects/wallet-service-reliable-integration/0001-design.md
@@ -122,6 +122,14 @@ The machine will transition to this state when a event is successfully processed
 **Transitions**:
   - The machine always transitions back to the `CONNECTED.idle` sub-state after performing the entry actions.
 
+**Services**:
+- The machine invokes (on entry) the `storeLastEvent` service, passing the event data and will only transition to `CONNECTED.idle` if the `event_id` was stored successfully
+
+**Transitions**:
+- If the service invocation completes successfully (`onDone`):
+    - **Target**: The machine transitions to the `CONNECTED.idle` sub-state.
+- If there's an error during the service invocation (`onError`):
+    - **Target**: The machine transitions to the `CONNECTED.error` sub-state.
 
 #### `CONNECTED.error`
 
@@ -140,6 +148,31 @@ This is a `final` state, meaning that no further transitions can occur from this
 
 
 # Guide-level explanation
+
+## Database design
+
+Aside from the wallet-service existing database, we need to store metadata related event processing:
+
+```sql
+CREATE TABLE sync_metadata (
+    id INT PRIMARY KEY,
+    last_event_id INT NOT NULL,
+    updated_at DATETIME DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
+);
+```
+
+`id` - This column will always be '1', the idea is to always have a single row in this table to keep track of the last event received
+`last_event_id` - This column will hold the last event that was successfully synced
+`updated_at` -
+
+This table should be created with the same migration mechanism the wallet-service uses. More about this in the monorepo section.
+
+## Monorepo
+
+We should merge both the wallet-service lambdas and the daemon in a single repository so they can share database migrations and we can have a better control over versions and database structures.
+
+I suggest we do this using [yarn workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/).
+
 
 ## Actors
 
@@ -164,6 +197,11 @@ This guard will if check the connected fullnode's peer-id is the same as the one
 This guard will if check the connected fullnode's Network is the same as the one set in the env `NETWORK`.
 
 ## Services
+
+#### storeLastEvent
+
+This service will store the last successful `event_id` on the `sync_metadata` table.
+
 
 #### handleVertexAccepted
 
--- a/projects/wallet-service-reliable-integration/0001-design.md
+++ b/projects/wallet-service-reliable-integration/0001-design.md
@@ -1,0 +1,205 @@
+## Summary
+
+The goal of this document is to describe a new method of syncing with a Hathor node, using the reliable integrations feature and replace the current sync daemon on our production services
+
+## Motivation
+
+At present, our wallet-service sync daemon operates by utilizing the fullnode's WebSocket and HTTP APIs to maintain synchronization with the fullnode. With this approach, we have to manually handle reorganizations, making it not very maintenance friendly as we have to deal with all potential edge cases that are already handled in the fullnode, again in the daemon.
+  
+It was precisely with such applications in mind that we developed our '[Reliable Integrations](https://github.com/HathorNetwork/rfcs/blob/master/projects/reliable-integration/0001-high-level-design.md)' feature. The concept behind this feature is to implement a reliable, integrated event management system. Through this system, a series of events would be dispatched in the exact order of their occurrence, ensuring that if we follow this exact series of events, the same state will be obtained
+
+## Reference-level explanation
+
+In a nutshell, when the reliable integrations mechanism proposes a built-in event management system for applications interacting with a full-node. This system ensures events are sent in the order they occur, assigns unique IDs to each event, and offers both REST API and WebSocket access for event queries.
+
+Given the synchronous nature of the event processing, I propose we design the daemon as a finite state machine as it can simplify the design by making it explicit which actions are valid in each state and how the system transitions between states.
+
+Here is the proposed state diagram for handling fullnode's events:
+
+```mermaid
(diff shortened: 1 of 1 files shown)


projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+In a nutshell, when the reliable integrations mechanism proposes a built-in event management system for applications interacting with a full-node. This system ensures events are sent in the order they occur, assigns unique IDs to each event, and offers both REST API and WebSocket access for event queries.
+
+Given the synchronous nature of the event processing, I propose we design the daemon as a finite state machine as it can simplify the design by making it explicit which actions are valid in each state and how the system transitions between states.
+
+Here is the proposed state diagram for handling fullnode's events:
+
+```mermaid
+stateDiagram-v2
+    [*] --> CONNECTING
+    CONNECTING --> CONNECTED: CONNECTED
```
msbrogli, line 22: After entering the `CONNECTED` state, you must confirm the following:

1) You are connected to the same full node as before (i.e., same `peer-id`).
2) This full node is connected to the correct network (e.g., mainnet).

If any of those are not true, you should go to state `error`.

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+## Potential challenges
+
+1. **Event Ordering and Consistency**: Given that events are sent in the order they occurred, it's crucial to ensure that the daemon maintains this order when processing events. This is especially important in scenarios like reorgs where the event order can significantly impact the wallet-service's state.
+2. **Event Volume and Scalability**: Depending on the full-node's activity, the number of events generated can be significant and considering that we are processing those events one at a time, we should make sure that we are doing this as efficiently as possible
+3. **Error Handling and Recovery**: There might be situations where the daemon encounters an error processing an event or loses connection to the full-node. We should make sure that we always have the latest successfully synced event id persisted and that we are rolling back changes that were unsuccessful to prevent invalid states
+
+
+# Guide-level explanation
```
msbrogli, line 138: After a connection is established, you are supposed to send the `id` of the next event you expect to receive. I couldn't find where you are persisting the next event `id` and the `peer-id` in case the sync daemon dies.

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+
+The goal of this document is to describe a new method of syncing with a Hathor node, using the reliable integrations feature and replace the current sync daemon on our production services
+
+# Motivation
+
+At present, our wallet-service sync daemon operates by utilizing the fullnode's WebSocket and HTTP APIs to maintain synchronization with the fullnode. With this approach, we have to manually handle reorganizations, making it not very maintenance friendly as we have to deal with all potential edge cases that are already handled in the fullnode, again in the daemon.
+  
+It was precisely with such applications in mind that we developed our '[Reliable Integrations](https://github.com/HathorNetwork/rfcs/blob/master/projects/reliable-integration/0001-high-level-design.md)' feature. The concept behind this feature is to implement a reliable, integrated event management system. Through this system, a series of events would be dispatched in the exact order of their occurrence, ensuring that if we follow this exact series of events, the same state will be obtained
+
+# Reference-level explanation
```
msbrogli, line 11: If I got it right, you are always running with a window of size 1. Is that right? If yes, please, make sure to add this piece of information to the design.

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+        idle --> idle: ACK
+        idle --> idle: '*'
+        
+        handlingMetadataChanged --> success: onDone
+        handlingMetadataChanged --> error: onError
+        
+        handlingVertexAccepted --> success: onDone
+        handlingVertexAccepted --> error: onError
+        
+        success --> idle
```
msbrogli, line 40: What if the connection is closed when you enter `success` but before sending the `ack`? Maybe you should store the `next_event_id` after you enter on success. Then, this id will be used on the `startStream` message if a connection is closed at this point.

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+#### WebSocketActor
+
+The `WebSocketActor` will initiate and maintain a connection with the full-node's WebSocket feed.
+
+We are using actors to have a bi-directional channel to the WebSocket feed from the machine, it will receive events and send those events to the machine, using the same type as received from the feed, so if new events are added in the future, they will be automatically emitted to the machine (and probably handled by the '\*' transition)
+
+
+## Services
```
msbrogli, line 153: If these methods are well tested, I feel it's ok to move forward with them. But I think there might a simpler way to handle these events. Let me explain.

Suppose events are group by `group_id`, and that you can create a new database transaction for each group, only committing after the last event of a group is processed. In this scenario, you can safely do the following:

1. When a group is started, you run a `START TRANSACTION` on the database.
2. Then, you process each transaction independently, updating the utxo table according to the event information. In other words, if an output is unspent, it goes into the utxo table. If its spent, you delete its related utxo. For clarity, you don't have to do anything related to inputs or parents.
2. You can process events in parallel (unless they are related to the same transaction).
3. After the whole group is processed, you can generate new addresses, unlock timelocked utxos and recalculate balance only once.
4. Finally, you `COMMIT` the changes to the database.

What do you think? Maybe we should consider doing something like this in the future since your current implementation is well tested.

msbrogli (changes requested):

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+
+*We already have those methods well tested in the wallet-service lambdas, we can just migrate them to the daemon.*
+
+#### handleMetadataChanged
+
+This service handles the `VERTEX_METADATA_CHANGED` event from the WebSocket feed, it will be emitted from the full-node on different situations:
+
+1. The full-node consensus mechanism changed the metadata of this vertex before accepting it
+2. This transaction has been confirmed by a block
+3. This transaction has been voided
```
glevco, line 181: As mentioned in another comment above, there are other situations that could trigger this event, but as I understand the situations on this list are the only ones we care about, right?

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+
+We need to detect and handle this event differently for each situation. We can detect the situation type by inspecting our own database and checking what changed.
+
+`(1)`:
+If we don't have this transaction on the database, it's the situation (1). We can just add it to the database by invoking the `handleVertexAccepted` service.
+
+`(2)`:
+We can do an `UPSERT` in the `transaction` database updating the `height` column. No balance re-calculation is needed here.
```
glevco, line 189: Why the `height`?

projects/wallet-service-reliable-integration/0001-design.md
andreabadesso: Sorry, added the service that will store the `event_id` in the `success` state

b2662064da15bf5ccc261f0fcef9c82aaa3ba49b

projects/wallet-service-reliable-integration/0001-design.md
andreabadesso: Since we only store the last `event_id` after it is successfully processed, when it gets in the `startStream` service again, it will send the correct `last_ack_event_id` and the sync will resume

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+  - **Actions**: The machine stores the event using the `storeEvent` action.
+  - **Target**: The machine transitions to the `handlingMetadataChanged` sub-state.
+- On receiving a `LOAD_STARTED` event:
+  - **Actions**: The machine stores the event with the `storeEvent` action.
+  - **Target**: The machine transitions to the `success` sub-state.
+- On receiving an `ACK` event:
+  - **Actions**: The machine sends an acknowledgment using the `sendAck` action.
+- On receiving any other event (represented by `'*'`):
+  - **Actions**: The machine stores the event with the `storeEvent` action.
+  - **Target**: The machine transitions to the `success` sub-state.
```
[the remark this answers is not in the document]
andreabadesso: Yes, the machine will receive "events" which could be either reliable integration events (as we are converting them to machine events) or other events

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+**Transitions**:
+- On receiving a `NEW_VERTEX_ACCEPTED` event:
+  - **Actions**: The machine stores the event using the `storeEvent` action.
+  - **Target**: The machine transitions to the `handlingVertexAccepted` sub-state.
+- On receiving a `VERTEX_METADATA_CHANGED` event:
+  - **Actions**: The machine stores the event using the `storeEvent` action.
+  - **Target**: The machine transitions to the `handlingMetadataChanged` sub-state.
+- On receiving a `LOAD_STARTED` event:
+  - **Actions**: The machine stores the event with the `storeEvent` action.
+  - **Target**: The machine transitions to the `success` sub-state.
```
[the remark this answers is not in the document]
andreabadesso: Just for clarity, I want to be clear about which events we are processing (and what is done when they're received), leaving all other events to the `'*'` case to make it future proof

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+#### `CONNECTED.handlingMetadataChanged`
+
+The machine will transition to this state to process a `NEW_METADATA_ACCEPTED` event.
+
+This event happens when the metadata of a transaction changes, this can happen for many reasons, but the ones we care about are:
+
+- The transaction has been confirmed by a block
+- The transaction was voided
```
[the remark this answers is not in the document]
andreabadesso: We handle tx outputs being spent on the `handlingVertexAccepted` service on the input side of the tx that spends it

projects/wallet-service-reliable-integration/0001-design.md
andreabadesso: Because that's what indicates to the wallet-service that the transaction has been confirmed by a block as we don't store metadata

projects/wallet-service-reliable-integration/0001-design.md
```diff
@@ -0,0 +1,205 @@
...
+
+Here is the proposed state diagram for handling fullnode's events:
+
+```mermaid
+stateDiagram-v2
+    [*] --> CONNECTING
+    CONNECTING --> CONNECTED: CONNECTED
+    state CONNECTED {
+        onEntry: startStream
+        on: DISCONNECT --> CONNECTING
```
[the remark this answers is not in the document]
andreabadesso: Updated it to be an image and a link to an interactive version:

019dd7f27eb0d081826c72b7bf7062fc37844c80

projects/wallet-service-reliable-integration/0001-design.md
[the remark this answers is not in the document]
glevco: Ok, thanks!

projects/wallet-service-reliable-integration/0001-design.md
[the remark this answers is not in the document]
glevco: But then, why not handle `LOAD_FINISHED` too, for example?

projects/wallet-service-reliable-integration/0001-design.md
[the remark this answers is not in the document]
glevco: Makes sense, thanks!

projects/wallet-service-reliable-integration/0001-design.md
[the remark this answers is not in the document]
andreabadesso: I forgot about it, done in 41d08656939b91e4b99f7fb4bd4222b9a93815aa

glevco (approved):

projects/wallet-service-reliable-integration/0001-design.md
andreabadesso: @msbrogli I added two future possibilities with the optimizations we discussed in our last call

projects/wallet-service-reliable-integration/0001-design.md
msbrogli: Great! Thanks!

projects/wallet-service-reliable-integration/0001-design.md
msbrogli: Great! Thanks!

projects/wallet-service-reliable-integration/0001-design.md
msbrogli: Great! Thanks!

projects/wallet-service-reliable-integration/0001-design.md
msbrogli: Great! Thanks for clarifying.

projects/wallet-service-reliable-integration/0001-design.md
msbrogli: Sounds great. Thanks!

msbrogli (approved):

PR #62, opened 2023-08-01
feat(feature-activation): phased testing

glevco:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/feat/feature-activation/phased-testing/projects/feature-activation/0004-phased-testing.md)

### Acceptance Criteria

- Implement a RFC describing a phased testing procedure for Feature Activation

### DIFF
--- a/projects/feature-activation/0004-phased-testing.md
+++ b/projects/feature-activation/0004-phased-testing.md
@@ -11,7 +11,7 @@ This RFC describes a phased procedure for testing the Feature Activation project
 # Motivation
 [motivation]: #motivation
 
-After all changes introduced by the previous Feature Activation RFCs are implemented, the mechanism should be tested with some "fake" features in both the testnet and mainnet environments, and only then we'll be able to use it with confidence in the mainnet to change real features like updating the maximum merkle path length, for example.
+After all changes introduced by the previous Feature Activation RFCs are implemented, the mechanism should be tested with some "dull" features in both the testnet and mainnet environments. Such features do not change any full node behavior and are used exclusively to test the Feature Activation mechanism itself. Only then we'll be able to use it with confidence in the mainnet to change real features like updating the maximum merkle path length, for example.
 
 # Guide-level explanation
 [Guide-level explanation]: #guide-level-explanation
@@ -42,7 +42,7 @@ In this phase, 3 NOP features are defined for testnet and each one of them under
 We'll be using 3 NOP features and define them in `testnet.yml`. We'll test different Feature Activation behavior by configuring different settings for each feature:
 
 1. `NOP_FEATURE_1` is expected to be signaled and activated through bit signaling. Therefore, its `signal_support_by_default` will be set to `True`, so testnet miners (the ones used in `tx-mining-service`) will receive block templates with their signal bit enabled. It'll also have a `minimum_activation_height`.
-2. `NOP_FEATURE_2` is expected to NOT be activated through bit signaling. Instead, it'll have `lock_in_on_timeout=True` and will be "forcefully" activated. (TODO: `MUST_SIGNAL` enforcement is not implemented yet. How do we set bits in testnet miners without using `signal_support_by_default`? What about in mainnet?)
+2. `NOP_FEATURE_2` is expected to NOT be activated through bit signaling. Instead, it'll have `lock_in_on_timeout=True` and will be "forcefully" activated.
 3. `NOP_FEATURE_3` is expected to FAIL. It'll have `signal_support_by_default=False` and `lock_in_on_timeout=False`.
 
 By doing this, we'll test different paths of the Feature Activation workflow. To define the configuration, we'll first choose some block height `N`, that is, the block height where Phase 1 will start. It can only be chosen when we actually implement this RFC. All other heights are relative to `N`. Here's the reference configuration:
@@ -184,7 +184,11 @@ And it's expected to reach its threshold and become active in 4 weeks (2 in `STA
 
 ## Phase 3
 
-Phase 3 can start as soon as Phase 1 ends, and it is completely analogous to Phase 1, except the NOP features will be configured on `mainnet.yml` instead of `testnet.yml`, and that `N` will be different.
+Phase 3 can start as soon as Phase 1 ends, and it is mostly analogous to Phase 1, minus a few exceptions:
+
+- The NOP features will be configured on `mainnet.yml` instead of `testnet.yml`.
+- `N` will be different.
+- We won't test `NOP_FEATURE_2`, that is, we won't test `lock_in_on_timeout=True` on mainnet, as we have no easy way of coordinating bit signals with miners during the `MUST_SIGNAL` phase.
 
 ## Conclusion
 
--- a/projects/feature-activation/0004-phased-testing.md
+++ b/projects/feature-activation/0004-phased-testing.md
@@ -1,0 +1,206 @@
+- Feature Name: Feature Activation Phased Testing
+- Start Date: 2023-08-01
+- Initial Document: [Feature Activation](https://docs.google.com/document/d/1IiFTVW1wH6ztSP_MnObYIucinOYd-MsJThprmsCIdDE/edit)
+- Author: Gabriel Levcovitz <<gabriel@hathor.network>>
+
+# Summary
+[summary]: #summary
+
+This RFC describes a phased procedure for testing the Feature Activation project.
+
+# Motivation
+[motivation]: #motivation
+
+After all changes introduced by the previous Feature Activation RFCs are implemented, the mechanism should be tested with some "fake" features in both the testnet and mainnet environments, and only then we'll be able to use it with confidence in the mainnet to change real features like updating the maximum merkle path length, for example.
+
+# Guide-level explanation
+[Guide-level explanation]: #guide-level-explanation
+
+The testing will consist of 3 phases:
+
+1. NOP features in testnet
+2. Maximum merkle path length update in testnet
+3. NOP features in mainnet
+
+Phase 1 will happen independently, and then Phase 2 and Phase 3 can happen concurrently. Note: updating the maximum merkle path length in mainnet is outside the scope of this project and will be described in a separate RFC.
+
+The NOP features used in Phases 1 and 3 are simply defined in the respective network's settings file, but their activation has no real effect in any computation of the full node, that is, no forks are created no matter whether those features are activated or not. Instead, their states will simply be logged by the full node.
+
+After Phase 1 is complete, Phase 2 will introduce a real change that creates a fork in testnet between nodes that recognize its feature activation, and nodes that do not. At the same time, Phase 3 can start. It is analogous to Phase 1, but on mainnet instead of testnet.
+
+# Reference-level explanation
+[Reference-level explanation]: #reference-level-explanation
+
+In this section, technical details are expanded for what was described above.
+
+## Phase 1
+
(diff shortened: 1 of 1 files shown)


projects/feature-activation/0004-phased-testing.md
```diff
@@ -0,0 +1,206 @@
...
+## Phase 1
+
+In this phase, 3 NOP features are defined for testnet and each one of them undergo their own feature activation process. Then, we validate that their states progress as expected, both by inspecting their values through the Explorer interfaces, and by inspecting logs via AWS CloudWatch queries.
+
+### NOP features configuration
+
+We'll be using 3 NOP features and define them in `testnet.yml`. We'll test different Feature Activation behavior by configuring different settings for each feature:
+
+1. `NOP_FEATURE_1` is expected to be signaled and activated through bit signaling. Therefore, its `signal_support_by_default` will be set to `True`, so testnet miners (the ones used in `tx-mining-service`) will receive block templates with their signal bit enabled. It'll also have a `minimum_activation_height`.
+2. `NOP_FEATURE_2` is expected to NOT be activated through bit signaling. Instead, it'll have `lock_in_on_timeout=True` and will be "forcefully" activated. (TODO: `MUST_SIGNAL` enforcement is not implemented yet. How do we set bits in testnet miners without using `signal_support_by_default`? What about in mainnet?)
```
glevco, line 45: When we implement mandatory signaling for the `MUST_SIGNAL` phase (which I should do before this Phased Testing), how will we set signal bits for our testnet miners? And how will we do it for mainnet, on Phase 3? We'll have to coordinate with ours partners, as using `signal_support_by_default=False` is part of the testing procedure. They'll have to manually set the bits on their side when `MUST_SIGNAL` is reached (or before).

projects/feature-activation/0004-phased-testing.md
```diff
@@ -0,0 +1,206 @@
...
+    if feature_service.is_feature_active(
+      block=block,
+      feature=Feature.INCREASE_MAX_MERKLE_PATH_LENGTH
+    ):
+        return NEW_MAX_MERKLE_PATH_LENGTH
+
+    return OLD_MAX_MERKLE_PATH_LENGTH
+```
+
+Where this function is actually defined will be left as an implementation detail. TODO: what's a good value for `NEW_MAX_MERKLE_PATH_LENGTH`?
```
glevco, line 163: @jansegre do we know what value we want to use for `NEW_MAX_MERKLE_PATH_LENGTH`?

projects/feature-activation/0004-phased-testing.md
```diff
@@ -0,0 +1,206 @@
...
+
+# Summary
+[summary]: #summary
+
+This RFC describes a phased procedure for testing the Feature Activation project.
+
+# Motivation
+[motivation]: #motivation
+
+After all changes introduced by the previous Feature Activation RFCs are implemented, the mechanism should be tested with some "fake" features in both the testnet and mainnet environments, and only then we'll be able to use it with confidence in the mainnet to change real features like updating the maximum merkle path length, for example.
```
msbrogli, line 14: I'd say "dull fatures" instead of "fake features", and maybe explain that dull features do not change any behavior and will be used exclusively to test the activation mechanism.

projects/feature-activation/0004-phased-testing.md
```diff
@@ -0,0 +1,206 @@
...
+
+# Guide-level explanation
+[Guide-level explanation]: #guide-level-explanation
+
+The testing will consist of 3 phases:
+
+1. NOP features in testnet
+2. Maximum merkle path length update in testnet
```
msbrogli, line 22: We shouldn't change it in the testnet if we don't plan to change it soon in the mainnet. I mean, testnet is a testing network that must mirror mainnet behavior.

projects/feature-activation/0004-phased-testing.md
msbrogli: I think such coordination is very complicated. We have to find a way to do it ourselves. Isn't it enough to do it on testnet? I mean, mainnet has a hash power that seems too high for us.

projects/feature-activation/0004-phased-testing.md
```diff
@@ -0,0 +1,206 @@
...
+    )
+```
+
+This logging will be removed after Phase 1 is complete.
+
+## Phase 2
+
+Phase 2 can start as soon as Phase 1 ends. Let `M` be a block height for the beginning of Phase 2. It'll define a new `INCREASE_MAX_MERKLE_PATH_LENGTH` feature that will actually affect how the full node verifies merged mining blocks.
+
+Currently, there's a `MAX_MERKLE_PATH_LENGTH: int = 12` constant in `aux_pow.py`. This constant will be replaced by a function call to `get_max_merkle_path_length()`, with its reference implementation as follows:
```
msbrogli, line 150: This refactor can already be implemented on a separate PR.

projects/feature-activation/0004-phased-testing.md
```diff
@@ -0,0 +1,206 @@
...
+
+And it's expected to reach its threshold and become active in 4 weeks (2 in `STARTED`, 2 in `LOCKED_IN`). We'll use the same validation mechanisms described in Phase 1 to make sure the feature states transition as expected. Also, we should configure a merged miner to mine blocks with a higher merkle path length, making sure its blocks are invalidated before the feature becomes `ACTIVE`, and validated after it does.
+
+## Phase 3
+
+Phase 3 can start as soon as Phase 1 ends, and it is completely analogous to Phase 1, except the NOP features will be configured on `mainnet.yml` instead of `testnet.yml`, and that `N` will be different.
+
+## Conclusion
+
+As described above, Phase 1 takes 6 weeks, and Phase 2 and Phase 3 can run concurrently, respectively for 6 and 4 weeks. This makes for a total Phased Testing duration of 12 weeks. If this duration is deemed too high, we can actually shift Phase 2 and 3 earlier, as they do not conflict with Phase 1 in any way. Another option to make it even faster, is decreasing the `EVALUATION_INTERVAL`, for example from 2 weeks to 1 week, only in testnet.
```
msbrogli, line 191: Should we test a feature deactivation after a rebase on testnet only? As the hash rate is so small, we should be able to do it easily. Or we can create a private network just to get this tested.

msbrogli (changes requested):

projects/feature-activation/0004-phased-testing.md
glevco: Indeed, but do you think this line should be removed from the text? Even if the function is created now in a separate PR, we'll need another PR later to use the Feature Activation mechanism inside the function

projects/feature-activation/0004-phased-testing.md
glevco: Did you mean reorg? If yes, then I agree it makes sense and I'll add it

projects/feature-activation/0004-phased-testing.md
glevco: I guess we do plan to change it soon on the mainnet, right? We could do it just after this testing process is complete and validated. If for any reason the testing fails on testnet and we need more time, we can always revert the change on testnet.

projects/feature-activation/0004-phased-testing.md
glevco: Makes sense. I agree we do it only on testnet. Changed in https://github.com/HathorNetwork/rfcs/pull/62/commits/2992fd5d8903b25c3a2605803d87b24e14d77603

projects/feature-activation/0004-phased-testing.md
glevco: Done in https://github.com/HathorNetwork/rfcs/pull/62/commits/2992fd5d8903b25c3a2605803d87b24e14d77603

msbrogli: Thanks!

projects/feature-activation/0004-phased-testing.md
msbrogli: In this case, deploying on testnet shouldn't be part of the rollout of the change on the maximum merkle path project? If it's simple enough, we can do everything on this project, so we will deploy this change on both the testnet and the mainnet.

projects/feature-activation/0004-phased-testing.md
msbrogli: You can keep this line here. It clearly explains how this change will occur.

projects/feature-activation/0004-phased-testing.md
msbrogli: Yes, reorg. My bad.

projects/feature-activation/0004-phased-testing.md
glevco: As discussed on meeting, the maximum merkle path length update for both testnet and mainnet will be described on a separate RFC. I left the original testnet text about it though, as reference.

projects/feature-activation/0004-phased-testing.md
glevco: As discussed on meeting, the maximum merkle path length update for both testnet and mainnet will be described on a separate RFC. I left the original testnet text about it though, as reference.

projects/feature-activation/0004-phased-testing.md
```diff
@@ -0,0 +1,209 @@
...
+        block_height=vertex.get_height(),
+        features_states=state_by_feature
+    )
+```
+
+This logging will be removed after Phase 1 is complete.
+
+### Phase 1b
+
+After the initial part of Phase 1 is complete, Phase 1b will start. On this phase, we'll force a reorg on testnet to undo the activation of `NOP_FEATURE_1`. We'll have to disable support of said feature in our miners and use a higher hash rate to generate more blocks than the original best chain, using some block before the 75% threshold is reached as the common block. After the procedure is complete, we must observe that `NOP_FEATURE_1` transitions back to `FAILED` instead of `ACTIVE`.
```
msbrogli, line 145: Should we run a separate private test network to test this part of the implementation? I mean, it's easy to cause a reorg on the testnet but maybe we are better off running this test in a private network.

msbrogli (approved):

projects/feature-activation/0004-phased-testing.md
msbrogli: Thanks!

projects/feature-activation/0004-phased-testing.md
msbrogli: Thanks!

projects/feature-activation/0004-phased-testing.md
msbrogli: Thanks!

projects/feature-activation/0004-phased-testing.md
msbrogli: Thanks!

projects/feature-activation/0004-phased-testing.md
glevco: We could, but I would also like to see this behavior happening in an environment as similar as possible to our mainnet, since we'll not test this on mainnet, of course.

If the testnet is too sensitive for this test, maybe we should have a new running network for development, like a dev-testnet, where we can freely do tests. What do you think? Maybe what I'm suggesting is just the same as a one-off private network...

msbrogli: I guess a dev network would be the same as a private network since we will control all peers, miners, and wallets connected to it.

Anyway, it's ok to use the public testnet for this. As we mine most of the blocks, we should be able to activate the feature as we wish. We should let the community know that we are running this test since they are supposed to upgrade their full nodes too.

PR #63, opened 2023-08-09
feat(feature-activation): add transactions RFC

glevco:


projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,182 @@
...
+
+1. After an `ACTIVE` block, all next blocks will also be `ACTIVE` (considering the same blockchain).
+2. Feature states for txs are only `ACTIVE` or `INACTIVE`.
+
+### Requirements
+
+Using the premises above, we must define a function that returns a state, given a transaction (note: we actually return multiple states for multiple features). That function must satisfy the following requirements:
+
+1. All txs that are received after (time-wise) an `ACTIVE` block in the best chain, must also be `ACTIVE`.
+2. When all txs are ordered by timestamp, there must not be an `INACTIVE` tx after an `ACTIVE` tx.
```
msbrogli, line 78: Timestamp might be tampered with. You might say that "there must not be an INACTIVE tx after an ACTIVE tx when the transactions are topologically sorted".

msbrogli (changes requested):

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Reward Lock
+
+Before detailing FATX, let's describe how the RL works in general terms. Then, we'll be able to observe the analogy and define FATX. As explained before, there are two main contexts for calculating RL (new vertices are separated into txs and blocks).
+
+#### Dealing with new txs
+
+1. A tx is received, and it spends the reward of a block. It is verified for reward lock. Then, there are two possibilities:
+   1. If the current best block's height IS NOT enough to unlock the reward, the tx is invalid and is rejected.
+   2. If the current best block's height IS enough to unlock the reward, the tx is valid and remains in the mempool.
```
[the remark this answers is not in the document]
glevco: This has been removed from the new iteration.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,182 @@
...
+
+#### Dealing with reorgs
+
+1. A reorg happens. Then, if the new best height is lower than the best height before the reorg,
+   1. Both txs that were already in the mempool, and confirmed txs that came back to the mempool, may have become invalid. For that, all txs in the mempool are re-verified (only for reward lock). If they're invalid, they're marked as such and removed from the mempool and the storage.
+
+#### Dealing with new blocks
+
+1. When a block is received, if one of the txs it confirms tries to spend a locked reward, the block is invalid and is rejected.
```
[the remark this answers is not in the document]
glevco: This has been removed from the new iteration.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,182 @@
...
+
+This rule is only for guaranteeing no rewards can be spent too early. In practice, it's impossible for such tx to be in the mempool, as it would have been invalidated before by the previous rules. TODO: Is this correct? Why does this rule exist, is it for compatibility with the previous reward lock mechanism?
```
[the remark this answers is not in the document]
glevco: This has been removed from the new iteration.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: This is the only comment that may still be relevant in the new iteration. Tampering of timestamps is accounted for in the RFC, please read the session dedicated to this.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,359 @@
...
+
+      This can be implemented by recursively hopping boundary blocks until
+      we find a block that is ACTIVE and has a parent that is LOCKED_IN.
+      """
+      raise NotImplementedError
+```
+
+## Dealing with reorgs
+
+Considering the definition above, a function must be called in the `ConsensusAlgorithm.update()` method that determines whether a reorg is invalid. If it is, the full node will **halt execution**.
```
jansegre, line 106: What exactly "halts execution" mean here? Would the fullnode's main process just exit with a non-zero code? What would we do with the database? Leave it in a non-consistent state? Wouldn't it be better to revert the re-org back to where it was before if a re-org is invalid?

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,359 @@
...
+
+With that in mind and going back all the way to the motivation of BIP 8 vs BIP 9 (as discussed in the original RFC defining Feature Activation for Blocks), we recall that block heights were used instead of block timestamps, mainly to prevent unexpected behavior when the hash rate is too volatile. However, using timestamps makes for a pretty straightforward solution for transactions. Combining that with the buffer idea, we can guarantee that an unwanted scenario is unlikely enough to consider it unrecoverable, forcing manual intervention. This buffer period can be as long as we want: the longer it is, the more unlikely is a manual intervention, but the delay for activating a feature for transactions increases. This tradeoff will be explored in sections below.
+
+Let's now define the solution: **for a feature, a transaction is considered `ACTIVE` if its timestamp is greater than one Evaluation Interval after the _first_ `ACTIVE` block's timestamp**. This will be detailed in the Reference-level section. The fact that timestamps are "mutable" and not signed will also be taken into account.
```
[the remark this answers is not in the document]
glevco: > Let me see if I understand. With this logic a feature that affects transactions cannot have a state on its own, only in the context of a given transaction a feature can have a state, right?

Correct.

> For example, consider that we want to add a field to a transaction, which affects parsing, we wouldn't be able to select a parser without knowing the tx timestamp, but we need a parser to get the tx timestamp, of course we can split the parser in 2 phases to solve this [...]

Indeed, this problem was described in the [Example - Releasing new Nano Contracts](https://github.com/HathorNetwork/rfcs/blob/feat/feature-activation/transactions/projects/feature-activation/0005-feature-activation-for-transactions.md#example---releasing-new-nano-contracts) section. Changing the schema of a transaction, like modifying, adding, or removing fields, would all have to be made through the creation of a new `TxVersion`, just like a new Nano Contract will be a new `TxVersion`.

The proposed solution is that the parser would simply succeed in deserializing both old and new `TxVersion`s, but we would have to add a new method in `verify()` that checks whether the tx's `TxVersion` is valid according to Feature Activation state.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: > What exactly "halts execution" mean here? Would the fullnode's main process just exit with a non-zero code?

Exactly. It's probably a good idea to add this to the text.

> What would we do with the database? Leave it in a non-consistent state?

Yes, it would work just like when we have a fail during loading, requiring the database to be reset. That's what I call "manual intervention".

> Wouldn't it be better to revert the re-org back to where it was before if a re-org is invalid?

Probably, but that would require us to implement the "database commit and rollback" mechanism in the storage which we currently don't have. I agree it would be nice, but I think that's a separate project since it touches critical code. The loading case, mentioned above, could also benefit from this.

But even if we do that, wouldn't the block that caused the reorg eventually be received again? Or you mean, reverting the reorg _and then_ halting the full node? I don't know if there would be many advantages in doing that, as the full node wouldn't be able to sync again anyway.

Also, we have to consider the likelihood of this ever happening, described in the [Likelihood of halting the full node](https://github.com/HathorNetwork/rfcs/blob/feat/feature-activation/transactions/projects/feature-activation/0005-feature-activation-for-transactions.md#likelihood-of-halting-the-full-node) section.

jansegre: Got it. I think it makes sense, if I have any considerations about the "likelihood" (because it might make sense to consider if it's possible for it to happen accidentally or the effort required to do it maliciously) I'll add a comment to that section.

projects/feature-activation/0005-feature-activation-for-transactions.md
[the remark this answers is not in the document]
jansegre: Got it, makes sense, I think it's a good solution to use the version field.

jansegre (dismissed):

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
+- Feature Name: Feature Activation for Transactions
+- Start Date: 2023-01-23
+- Initial Document: [Feature Activation](https://docs.google.com/document/d/1IiFTVW1wH6ztSP_MnObYIucinOYd-MsJThprmsCIdDE/edit)
+- Author: Gabriel Levcovitz <<gabriel@hathor.network>>
+
+# Summary
+[summary]: #summary
+
+This document describes a way to use the Feature Activation process with Transactions, complementing the existing implementation for Blocks. The original [Feature Activation for Blocks RFC](https://github.com/HathorNetwork/rfcs/blob/master/projects/feature-activation/0001-feature-activation-for-blocks.md) can be read for more detailed information on what the Feature Activation process is, why it is necessary, and how it works.
```
msbrogli, line 9: I guess we should add that this design does not address some types of features such as removing OPCODES. The solution for these cases will be addressed in the future.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
...
+
+Then, we arrived at the solution proposed in this document, which has most of the same advantages from the parent block solution, however it does not require any breaking change whatsoever. Here it is: **for a feature, a transaction is considered `ACTIVE` if its closest block, that is, its ancestor block with the greatest height, is `ACTIVE`**. This will be detailed in the Reference-level section.
+
+In other words, transactions today can already point to blocks via their inputs. Only a small portion of transactions spend from blocks, though. The idea is then to propagate this indirect block dependency to all transactions. This can be achieved by simply creating a new metadata that is set when a transaction is received from the network. The calculation is deterministic, equal for all peers, and with constant execution complexity. Also, reorgs are handled automatically by existing code, and no change has to be made in critical components such as the consensus.
```
msbrogli, line 32: I guess this section has some historical background but it does not explain the solution in enough depth. I think we must name the new metadata field and explain what it means giving at least one example. And then explain how this metadata will be used to enable features for transactions.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
...
+Using the premises above, we must define a function that returns a state, given a transaction (note: we actually return multiple states for multiple features). That function must satisfy the following requirements:
+
+1. A while after one `ACTIVE` block in the best chain, it must be possible to create `ACTIVE` transactions.
+2. A transaction that has an `ACTIVE` transaction as a dependency, is also `ACTIVE`.
+
+### Definition
+
+Repeated here from the Guide-level section:
+
+**For a feature, a transaction is considered `ACTIVE` if its closest block, that is, its ancestor block with the greatest height, is `ACTIVE`.**
```
msbrogli, line 59: I guess we should be careful with the "closest block" expression since it might be misleading. Maybe use the expression `closest ancestor block` or `closest pointed block`.  I also think that you should expand the explanation to clarify that the transaction itself might be spending a mining reward which will make this block a candidate. I believe that two or three examples should clarify all cases.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
...
+    """
+    from hathor.transaction import Block, Transaction
+    if isinstance(self, Block):
+        return
+    assert isinstance(self, Transaction)
+    assert self.storage is not None
+    metadata = self.get_metadata()
+
+    if self.is_genesis:
+        metadata.closest_block = self._settings.GENESIS_BLOCK_HASH
```
msbrogli, line 104: Should we point out this edge case in the description?

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
...
+        vertex = self.storage.get_transaction(vertex_id)
+        vertex_meta = vertex.get_metadata()
+        this_closest_block: Block
+
+        if isinstance(vertex, Block):
+            assert vertex_meta.closest_block is None
+            this_closest_block = vertex
+        elif isinstance(vertex, Transaction):
+            this_closest_block_id = (
+                self._settings.GENESIS_BLOCK_HASH if vertex.is_genesis else not_none(vertex_meta.closest_block)
```
msbrogli, line 120: I guess you have already handled this edge case above. So here you can safely just do `this_closest_block_id = not_none(vertex_meta.closest_block)`.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
...
+            closest_block = this_closest_block
+
+    assert closest_block is not None
+    assert closest_block.hash is not None
+    metadata.closest_block = closest_block.hash
+```
+
+A migration will also be necessary to populate this new metadata for existing transactions.
+
+## Example - Releasing new Nano Contracts
```
msbrogli, line 136: Move the example to the Guide-level explanation and make sure the guide-level explanation provide enough details to one can understand the example.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
...
+
+Then, a new method must be added in the verification phase of `BaseTransaction`, to assert that the deserialized `TxVersion` is valid. In that method, `is_feature_active_for_transaction()` would be called, and depending on the state returned, the `TxVersion` representing the new Nano Contract would be accepted or not.
+
+This demonstrates that there are two possible known use cases for Feature Activation for Transactions:
+
+1. Changing behavior in transaction validation. In this case, the usage is straight forward, simply a call to `FeatureService.is_feature_active_for_transaction()` to check whether the new feature is `ACTIVE`.
+2. Changing behavior in transaction deserialization. In this case, it may be necessary to create a new transaction validation method that verifies the validity of using a new deserialization feature.
+
+## Mutability of transaction parents
```
msbrogli, line 149: Same here: move to the Guide-level explanation.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
...
+
+In other words, if a third-party tries to manipulate a transaction such that its activation is toggled, the copied transaction will only remain valid if the usage of a feature doesn't affect the validity of that transaction in the first place.
+
+Therefore, the mutability of transaction parents is not an issue for the solution proposed in this document.
+
+## Selection of transaction parents
```
msbrogli, line 159: Same here: move to the Guide-level explanation.

projects/feature-activation/0005-feature-activation-for-transactions.md
```diff
@@ -0,0 +1,196 @@
...
+
+This means that it's not possible to enforce a new feature on old transactions using this mechanism. For example, it would not be possible to remove the validity of an existing opcode, as someone could simply create a transaction spending and confirming transactions from before this removal was activated. For the known use cases, this is not an issue. Every planned feature at this moment can be activated through the solution in this document. A solution for removing existing features will be defined in a separate RFC, in the future, if necessary.
+
+# Rationale and alternatives
+[Rationale and alternatives]: #rationale-and-alternatives
+
+This RFC passed through a lot of completely different idea iterations before arriving at this solution. Please see the [iterations directory](./0005-iterations) for more information.
+
+# Prior art
```
msbrogli, line 183: It's missing a Future possibilities where we need to add the ability to enable more advanced features such as removing an OPCODE.

msbrogli (changes requested):

projects/feature-activation/0005-feature-activation-for-transactions.md
jansegre: I think the proposed mechanism is very good, especially for its simplicity, but I also think "closest block" is not a great expression. I'll think of some alternatives and suggest here.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: This is mentioned in the Drawbacks section.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: This is detailed in the Reference-level section and an example is provided in "Example - Releasing new Nano Contracts"

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: This is just reference code and in the actual implementation this was removed.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: This was renamed to `closest_ancestor_block` in the implementation, and I just updated it here.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: Done!

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: Done, I updated this with the actual implementation which does not include this case.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: Done!

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: I don't think this should be moved to the guide-level, this is just a consideration on whether the solution has this problem or not.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: I also think that this is just an implementation detail that should be in the reference level, not guide level.

projects/feature-activation/0005-feature-activation-for-transactions.md
glevco: Done!

projects/feature-activation/0005-feature-activation-for-transactions.md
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: Ok. Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: Ok. Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: Ok. Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: Ok. Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: Thanks!

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: We must at least mention that the parent selection will be critical to enable new features. Even though the DAG width is usually low, it might happen to be high and transactions might fail due to a poor parent selection. The low level should cover on the details of how this selection should take place. This part is critical for wallets.

projects/feature-activation/0005-feature-activation-for-transactions.md
msbrogli: Ok.

msbrogli (approved):

PR #65, opened 2023-08-16
design: dev settings in the wallet-mobile

alexruzenhack:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/e470656eb9706c17e4702f6eb02969b2d32c44e5/projects/hathor-wallet-mobile/0001-dev-settings.md)

## Acceptance criteria

- Design of the Dev Settings in the wallet-mobile
  - Add the possibility to change the network settings
  - Comply with usability heuristics to avoid errors and mitigate risks
  - Change the UI if the wallet uses a custom network settings

## Task break-down
[test-break-down]: #test-break-down

| Task                                                                        | Dev/days |
| --------------------------------------------------------------------------- | -------- |
| Add the pre-defined network settings                                        | 0.1      |
| Add default network settings                                                | 0.1      |
| Implement the networkSettings saga                                          | 0.5      |
| Implement the Alert UI                                                      | 0.2      |
| Refactor the wallet saga                                                    | 1        |
| Refactor the public explorer component                                      | 0.1      |
| Refactor the Settings page to contain the "General Settings" and the "Dev Settings" | 0.5      |
| Implement the "Risk disclaimer" page                                        | 0.2      |
| Implement the "Pre settings" page                                           | 0.5      |
| Implement the "Custom settings" page                                        | 0.5      |


trondbjoroy (commented):
Should we have a Cancel button on the Pre settings page, which takes the user back to Settings page?

And on the Custom settings page, should we have a "Back" button, which takes the user back to the Pre settings page?

Alternatively, instead of different buttons on these pages, perhaps it would be better to add a "left-arrow" at the top of the pages, which would navigate one page back.

Later, we should also do a project to make the network settings flow identical in both the mobile and desktop wallets.

alexruzenhack:
> Alternatively, instead of different buttons on these pages, perhaps it would be better to add a "left-arrow" at the top of the pages, which would navigate one page back.

Sure! All the pages will have this "left-arrow" because this is a standard in the wallet mobile. The images were just schemes to illustrate the idea.

> Later, we should also do a project to make the network settings flow identical in both the mobile and desktop wallets.

Ok. I will add it in the future work session.

pedroferreira1:
I like the design and it's approved for me, but I have just some thoughts:

- I don't like the message "custom network" as red, it seems error but it's not, I prefer yellow, as the warning color.
- Which features need the wallet service? Can we disable those features if the user don't fill the ws inputs?

andreabadesso (approved):

pedroferreira1 (approved):

PR #67, opened 2023-08-22
feat: Headless HSM - Client side

tuliomir:
( See [rendered](https://github.com/HathorNetwork/rfcs/blob/feat/headless-hsm/projects/hathor-wallet-headless/hsm-integration.md) )

Design of the integration between the Headless Wallet app and the HSM

### DIFF
--- a/projects/hathor-wallet-headless/hsm-integration.md
+++ b/projects/hathor-wallet-headless/hsm-integration.md
@@ -4,6 +4,9 @@
 - Hathor Issue:
 - Author: Tulio Vieira tulio@hathor.network
 
+- Say that there will be a need to change the lib if we want to use simple_send_tx
+- Say that for now only the tx proposal routes will be integrated with the HSM
+
 # Summary
 [summary]: #summary
 
@@ -117,6 +120,7 @@ A proof-of concept application was partially developed in C as a bridge between 
 This means that headless applications that are supposed to interact with the HSM should copy the compiled executables of the *HSM Client* to their environment before or during operation.
 
 ## Headless implementation
+[headless-implementation]: #headless-implementation
 The headless application will be responsible for calling the *HSM Client* directly.
 
 The connection authentication credentials will be stored securely on the environment variables, allowing the leverage of other security best practices to avoid credential leaks.
@@ -151,7 +155,9 @@ If the generated `xPub` is not valid, an error is thrown to the user indicating 
 | :--- |
 | The HSM API to retrieve the `xPub`  from a generated `BIP32` wallet is still under development. An initial estimate from the Dinamo team is for this to be available on September. A degree of development uncertainty can be found when this API is available. |
 
-Once the wallet is started successfully, any request to fetch signatures or to sign transactions will send the data to be signed to the *HSM Client*. This can be based on the desktop hardware wallet [`ledger/sendTx()` function](https://github.com/HathorNetwork/hathor-wallet/blob/ef57015015375a477cffd72baf62f4e14baf541a/src/utils/ledger.js#L126-L166).
+Once the wallet is started successfully, the signature of HSM inputs can be made through a `Tx Proposal`, using the Atomic Swap API. The `wallet/atomic-swap/get-my-signatures` route will be modified to identify that the wallet is related to an HSM and will send the data to be signed to the *HSM Client*. This can be based on the desktop hardware wallet [`ledger/sendTx()` function](https://github.com/HathorNetwork/hathor-wallet/blob/ef57015015375a477cffd72baf62f4e14baf541a/src/utils/ledger.js#L126-L166).
+
+_Note:_ For this initial version, only this route will be adapted for using the hardware wallet, since it's the only one that has an implementation close enough to the Desktop Ledger. The other endpoints, such as `wallet/simple_send_tx` will require more complex refactorings and will be discussed on the _Future Possibilites_ section.
 
 | ⚠ Incomplete PoC - sign tx |
 | :--- |
@@ -280,6 +286,13 @@ This RFC also aims to confirm the proposed solution as superior choice in cost-b
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
+## Full integration on all headless endpoints
+As explained in the [headless implementation section](#headless-implementation), only the atomic swap `get-my-signatures` endpoint was adapted to interact with the HSM.
+
+Adapting all endpoints to use the HSM would require more profound refactorings. For example, the `wallet/simple-send-tx` endpoint uses a complete abstraction from the Wallet Lib to make all the transaction information and sign within a single lib call.
+
+A discussion would be necessary to identify the best way to implement this: either through a refactor of the lib code itself, receiving the HSM information via optional parameters, or rethinking the implementation of these methods on the headless to avoid the simplifications offered by the lib.
+
 ## Optimizations on the *HSM Client*
 As mentioned on the [alternatives](#rationale-and-alternatives) section, many improvements could be implemented on the _HSM Client_, including:
 - Developing a full daemon application to run it
@@ -287,6 +300,17 @@ As mentioned on the [alternatives](#rationale-and-alternatives) section, many im
 
 Future implementations of this solution could also see the HSM wallet usage on the Desktop wallet, for a friendlier GUI to the end user.
 
+## Direct access to C libraries
+Some npm libs such as [`node-gyp`](https://www.npmjs.com/package/node-gyp) indicate the potential to directly access the HSM C libraries without having to compile executables for the Headless environment separately.
+
+There are, however uncertainties about their usage and added complexity of using them on an existing application. So they need extra investigation before being considered an option.
+
 # Task Breakdown
-| 🚧 Work in Progress |
-| --- |
+| Task                                                                       | Effort | Pending PoC |
+|----------------------------------------------------------------------------|:------:|:-----------:|
+| Validate `/start` requests with `hsmWallets` config                        |  0.2   |             |
+| Implement call to _HSM Client_ `getXPub` on wallet start                   |  0.3   |      ⚠      |
+| Unit test success and errors with `getXPub` call                           |  0.2   |             |
+| Mock `getXPub` result and start a wallet with it                           |  0.3   |             |
+| Implement call to _HSM Client_ `signData` on `/get-my-signatures` endpoint |  0.8   |      ⚠      |
+
--- a/projects/hathor-wallet-headless/hsm-integration.md
+++ b/projects/hathor-wallet-headless/hsm-integration.md
@@ -188,7 +188,7 @@ The main difference between the Ledger and the HSM is that here there is no need
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-The only question left is if our Wallet Lib is able to correctly sign a transaction with the HSM in its current state, or if changes will have to be made to it to allow this integration. The 2nd phase will be dependent on this answer.
+The only question left is if our application is able to correctly sign a transaction with the HSM in its current state, or if changes will have to be made to the Hathor Lib or the Dinamo-Hsm lib to allow this integration. The 2nd phase will be dependent on this answer, and it is being discussed on [internal issue 205](https://github.com/HathorNetwork/internal-issues/issues/205).
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
--- a/projects/hathor-wallet-headless/hsm-integration.md
+++ b/projects/hathor-wallet-headless/hsm-integration.md
@@ -102,61 +102,40 @@ Eventual network failures when trying to communicate with the HSM are retried an
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-This project has three main phases that need implementing:
-1. The communication with the HSM itself. We will call this the *HSM Client* throughout this document.
-2. Adapting the Headless wallet to interact with this client and use the HSM for cryptographic operations
+This project has two main phases that need implementing:
+1. Starting a `read only wallet` that fetches all relevant data from the HSM but can not sign a transaction.
+2. Implementing a signature integration with the HSM.
 
-## The HSM Client
-The main library offered by Dinamo to interact with the HSM device [is available in C language](https://manual.dinamonetworks.io/c/index.html).
-
-A proof-of concept application was partially developed in C as a bridge between the Wallet Lib javascript environment and the C environment. A dedicated design will be written about this interface, but in short:
-- The C code expects to be called from the operating system. In Node.js terms this means using [`child_process.spawn()`](https://nodejs.org/docs/latest-v18.x/api/child_process.html#child_processspawncommand-args-options)
-- Parameters such as the derivation path and input data to be signed should be passed through `stdin`, not the `argv` command parameters. This avoids leaking sensitive data to the public command line history of the OS
-- Responses are sent through `stdout` and read using the child process `stdout.on('data')` event listener
-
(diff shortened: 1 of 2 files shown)


projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,313 @@
...
+HSM_DEVICE_PASSWORD="my_password"
+```
+
+A configuration attribute will be added to the `config.js` file, similar to the existing `seeds` attribute, mapping the headless `walletId`s to HSM object key names.
+```js
+// On config.js
+const hsmWallets = {
+  'headless-wallet-id': 'HSM_KEY_NAME'
+}
+```
```
r4mmer, line 137: When running on docker (the intended way for headless) the config.js loads the configuration from the environment variables, so we should also show how the environment variables fot this config will be done.

Another point to be sure is that we have a system to reload configuration with the wallet running.
We need to map on the settings module how the wallet will react if this configuration changes.

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,313 @@
...
+
+The connection authentication credentials will be stored securely on the environment variables, allowing the leverage of other security best practices to avoid credential leaks.
+```sh
+# On env variables
+HSM_DEVICE_HOST="192.168.0.1"
+HSM_DEVICE_USERNAME="my_username"
+HSM_DEVICE_PASSWORD="my_password"
+```
+
+A configuration attribute will be added to the `config.js` file, similar to the existing `seeds` attribute, mapping the headless `walletId`s to HSM object key names.
```
r4mmer, line 131: From what I understand, we will start the wallet and the wallet will identify that this wallet-id is supposed to be stated as an HSM client.

I think that there is a better way to make this, I think we should have a dedicated hsm path, and a dedicated start method for hsm wallets.
So it would be something like `/hsm/start` which receives the `wallet-id` (and I think there are no other parameters for hsm wallets) and handles calling the HSM api and starting the lib.
What do you think?

Obs: We could even pass the username and password as post parameters and save them on a map for wallet-id to hsm config, so we would have no hsm config in the env which is a little bit safer.

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,313 @@
...
+
+Once the wallet is started successfully, the signature of HSM inputs can be made through a `Tx Proposal`, using the Atomic Swap API. The `wallet/atomic-swap/get-my-signatures` route will be modified to identify that the wallet is related to an HSM and will send the data to be signed to the *HSM Client*. This can be based on the desktop hardware wallet [`ledger/sendTx()` function](https://github.com/HathorNetwork/hathor-wallet/blob/ef57015015375a477cffd72baf62f4e14baf541a/src/utils/ledger.js#L126-L166).
+
+_Note:_ For this initial version, only this route will be adapted for using the hardware wallet, since it's the only one that has an implementation close enough to the Desktop Ledger. The other endpoints, such as `wallet/simple_send_tx` will require more complex refactorings and will be discussed on the _Future Possibilites_ section.
+
+| ⚠ Incomplete PoC - sign tx                                                                           |
+|:-----------------------------------------------------------------------------------------------------|
+| This step was also not fully implemented on the PoC and can add uncertainty to the development phase |
```
r4mmer, line 157: I think we should remove this from warning from the design

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,313 @@
...
+[headless-implementation]: #headless-implementation
+The headless application will be responsible for calling the *HSM Client* directly.
+
+The connection authentication credentials will be stored securely on the environment variables, allowing the leverage of other security best practices to avoid credential leaks.
+```sh
+# On env variables
+HSM_DEVICE_HOST="192.168.0.1"
+HSM_DEVICE_USERNAME="my_username"
+HSM_DEVICE_PASSWORD="my_password"
+```
```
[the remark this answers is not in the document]
tuliomir: A javascript library was made available by the third party, removing the need for a HSM Bridge solution.

The design was updated to reflect this.

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: Good point! The description was simplified to accomodate the new Settings module and soft reload feature.

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: Removed.

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: Good suggestions. Incorporated them both on the main design and "Future Possibilities" sections.

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,256 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+The only question left is if our application is able to correctly sign a transaction with the HSM in its current state, or if changes will have to be made to the Hathor Lib or the Dinamo-Hsm lib to allow this integration. The 2nd phase will be dependent on this answer, and it is being discussed on [internal issue 205](https://github.com/HathorNetwork/internal-issues/issues/205).
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+## Full integration on all headless endpoints
```
r4mmer, line 196: I understand that full compatibility is not the goal of this project but I think we should at least describe what the options of implementations are. This way the discussion of which implementation is better will be faster.

What do you think?

r4mmer (approved):

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: Good idea. I added a draft proposal to adapt the `prepareTransaction` with this purpose on [36c319b](https://github.com/HathorNetwork/rfcs/pull/67/commits/36c319bb42dc11b8f853d2d549bc04ebf837eb87).

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,280 @@
...
+For the Wallet Lib this means that the following operations are executed within the HSM environment:
+- Generation of a `BIP-32 xPriv`
+- Generation of an `xPub` for this `xPriv`
+- Derivation of all child private keys for this `xPriv`
+- Signing transaction inputs
+
+### The HSM Environment
+The full documentation of the HSM and its API can be found on the [Dinamo HSM docs GitBook](https://docs.dinamonetworks.com/). In short, the architecture of this device from our client standpoint consists of:
+- A web server that provides the API endpoints
+- Each connection to this server consists a **session**, and is authenticated using a username and password
```
pedroferreira1, line 43: What happens if the wallet loses connection to this session while using it? Is there any way of listening to this information? Or do we need to send a request to it to check?

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,280 @@
...
+
+If the generated `xPub` is not valid, an error is thrown to the user indicating that the designated HSM key does not contain a valid `BIP32 xPriv` and is unsuitable for using as a wallet.
+
+Once the wallet is started successfully, all other operations are executed in a similar fashion to the current headless implementation of a read-only wallet.
+
+## Signature integration - Phase 2
+The signature of HSM inputs will be made through a `Tx Proposal`, using the Atomic Swap API. The `wallet/atomic-swap/get-my-signatures` route will be modified to identify that the wallet is related to an HSM and will send the data to be signed to the *HSM Client*. This can be based on the desktop hardware wallet [`ledger/sendTx()` function](https://github.com/HathorNetwork/hathor-wallet/blob/ef57015015375a477cffd72baf62f4e14baf541a/src/utils/ledger.js#L126-L166).
```
pedroferreira1, line 152: Why do you need to use the `atomic-swap` APIs? We have nowadays the API in the `tx-proposal` routes to build a tx, add signatures to a tx. Do you need anything else? I think you will create a new API in the `hsm` router group to get this signatures from the hsm device, but after you can use the ones from the `tx-proposal` routes.

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,280 @@
...
+
+Only this route will be adapted for using the hardware wallet, since it's the only one that has an implementation close enough to the Desktop Ledger. The other signature endpoints, such as `wallet/simple_send_tx` would require more complex refactorings and will be discussed on the _Future Possibilites_ section.
+
+## Creating a BIP32 on the HSM
+This operation should be done via script, as it is the most critical operation on the HSM within this scope.
+
+As such, a new `scripts/createHSMKey.js` file will be created, offering a way to quickly generate a new BIP32 wallet on a developer machine. This created key name should be inserted on the `config.js` file along with a `walletId` for it. Other wallet generation possibilities are discussed on the _Alternatives_ section of this document.
```
pedroferreira1, line 159: I didn't understand this. What will be added to the `config.js`? Which field and why?

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,280 @@
...
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+## Full integration on all headless endpoints
+As explained in the [headless implementation section](#headless-implementation), only the atomic swap `get-my-signatures` endpoint was adapted to interact with the HSM.
+
+Adapting all endpoints to use the HSM would require more profound refactorings, more specifically the [utils/transaction#prepareTransaction()](https://github.com/HathorNetwork/hathor-wallet-lib/blob/d3dbe159ac121eb67986ed8561cdacead8fc9fe8/src/utils/transaction.ts#L485-L502) method. From the perspective of the headless wallet, all signatures are executed through this method.
+
+Below are the endpoints that would need to be integrated with the HSM ( [link to the router file](https://github.com/HathorNetwork/hathor-wallet-headless/blob/1fc34b6f12a349478d9656b14933e6db45115a90/src/routes/wallet/wallet.routes.js) for conference ):
```
pedroferreira1, line 201: You don't need to add specific support for each of these endpoints. We can add support to build any type of tx, then the signature/mining/push of the tx is the same for all of them.

projects/hathor-wallet-headless/hsm-integration.md
```diff
@@ -0,0 +1,280 @@
...
+    - [sendTransaction](https://github.com/HathorNetwork/hathor-wallet-lib/blob/a0e20f6aa9c07461d7784e3133a6fedc4b72fabf/src/new/wallet.js#L1185)
+
+A new function could be added to the wallet storage, allowing the `prepareTransaction` to identify the wallet is linked to an HSM device. From this point, the function itself could retrieve the connection credentials from the config file and call the HSM lib directly, executing the signature.
+
+Extra attention would be needed to ensure that thrown exceptions from the HSM lib would be correctly handled by the Wallet Lib, and that the headless end user would receive a friendly error message.
+
+## Avoiding the use of the `xPub`
+There is the possibility of some partner needing a wallet with a higher level of security, avoiding even the export of the `xPub` data.
+
+Should this become a requirement, another path to implement this integration would be to modify the Wallet Lib, allowing the starting of a wallet without a way to derive public keys internally.
```
pedroferreira1, line 230: The wallet nowadays is ready to start a read only wallet from a list of addresses only. I think this would be enough for this requirement, we would just need to implement this support in the headless, not in the lib.

projects/hathor-wallet-headless/hsm-integration.md
pedroferreira1: I think now I understand after reading the code. You never keep the session opened, right? You send the request to the HSM, then disconnects, right? If that's the case, it's good to be clear here in the design (or I might have missed this information)

tuliomir: I added some clarifications about the singleton nature of the HSM connections on [62b08e6](https://github.com/HathorNetwork/rfcs/pull/67/commits/62b08e6930f53ccbb681c30ced836f7554354c35).

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: Previous iterations of this design had the keys set on the `config.js`, this is a leftover from these older versions.
Fixed on [12985ed](https://github.com/HathorNetwork/rfcs/pull/67/commits/12985ed8320681ac512326809f776bee792fca88).

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: Reworded this section on [66c6c50](https://github.com/HathorNetwork/rfcs/pull/67/commits/66c6c50949a9f15e81867c416b5dca57885af1bb). However, I could not find when did this feature was implemented to add as a reference on the document. Do you know which PR added it?

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: The final conclusion of this section of the document was also that: we could add support to a specific method on the wallet lib and compatibility with HSM signatures would become automatic.

On [ce01489](https://github.com/HathorNetwork/rfcs/pull/67/commits/ce01489eec9c6976213f825171471b06764b8c04) removed the section that contained the research part and left only the final conclusion of where the adaptation will be needed on the wallet lib.

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: I tried to find an endpoint on `tx-proposal` routes that allowed the signature of transactions, but I couldn't find any.

The `p2sh` routes have a way to generate the signatures, but regular `p2pkh` transactions only have routes to add existing signatures that were already calculated to the transaction ( see [headless docs site](https://wallet-headless-api.docs.hathor.network/#operation/createTxProposal) ).

Maybe there is something I'm missing here?

projects/hathor-wallet-headless/hsm-integration.md
pedroferreira1: This was a long time ago. You can reference the line with the attribute of the constructor in the `HathorWallet` class.

https://github.com/HathorNetwork/hathor-wallet-lib/blob/v1.1.0/src/new/wallet.js#L107

projects/hathor-wallet-headless/hsm-integration.md
pedroferreira1: We really don't have an endpoint to generated the signatures in the `tx-proposal` routes. I think we should create one instead of using the `atomic-swap` routes for something that isn't an atomic swap.

tuliomir: Makes sense. Added the reference and explanation for the new `hsm/get-my-signatures` route on [ac26cfa](https://github.com/HathorNetwork/rfcs/pull/67/commits/ac26cfa5c7a8db84e58eafe98bff808b502c0299)

projects/hathor-wallet-headless/hsm-integration.md
tuliomir: I incorporated our ideas on https://github.com/HathorNetwork/hathor-wallet-headless/issues/393 and made a reference to them on the design on commit [c8a5afd](https://github.com/HathorNetwork/rfcs/pull/67/commits/c8a5afd14886ae8f453c17914f5602279cba89ac)

pedroferreira1 (approved):

r4mmer (approved):

PR #68, opened 2023-08-25
feat: Headless HSM - C Bridge

tuliomir:
( See [rendered](https://github.com/HathorNetwork/rfcs/blob/feat/headless-hsm-bridge/projects/hathor-wallet-headless/hsm-c-bridge.md) )

Design of the integration between the Dinamo C Library and a Node.js client application used by the Headless Wallet.

### DIFF
--- a/projects/hathor-wallet-headless/hsm-c-bridge.md
+++ b/projects/hathor-wallet-headless/hsm-c-bridge.md
@@ -79,89 +79,89 @@ Sample code for signing a transaction:
 
 int main(int argc, char **argv)
 {
-	// Return value of the latest operation
-	int nRet = 0;
-	struct AUTH_PWD authPwd;
-	HSESSIONCTX hSession = NULL;
+  // Return value of the latest operation
+  int nRet = 0;
+  struct AUTH_PWD authPwd;
+  HSESSIONCTX hSession = NULL;
 
-	// XPriv Key name
-	char szId[] = HTR_TEST_XPRIV_KEYNAME;
+  // XPriv Key name
+  char szId[] = HTR_TEST_XPRIV_KEYNAME;
 
-	BYTE ver = DN_BCHAIN_VER_WIF_TEST_NET;
+  BYTE ver = DN_BCHAIN_VER_WIF_TEST_NET;
 
-	// Address to derive child priv key from
-	int nChild = 0;
+  // Address to derive child priv key from
+  int nChild = 0;
 
-	// Child Priv Key Name
-	char szDest[MAX_OBJ_ID_FQN_LEN] = "HTR_CHILD_KEY";
-	DN_BCHAIN_KEY_INFO pbKeyInfo;
+  // Child Priv Key Name
+  char szDest[MAX_OBJ_ID_FQN_LEN] = "HTR_CHILD_KEY";
+  DN_BCHAIN_KEY_INFO pbKeyInfo;
 
-	// Message buffer
-	BYTE pbMsg[1024] = {0};
+  // Message buffer
+  BYTE pbMsg[1024] = {0};
 
-	// Output encoding buffer
-	char pbOutputEncode[1024 * 2 + 1];
-	DWORD dwOutputLength = sizeof(pbOutputEncode);
+  // Output encoding buffer
+  char pbOutputEncode[1024 * 2 + 1];
+  DWORD dwOutputLength = sizeof(pbOutputEncode);
 
-	// Hash buffer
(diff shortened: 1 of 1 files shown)


projects/hathor-wallet-headless/hsm-c-bridge.md
```diff
@@ -0,0 +1,327 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+This bridge is necessary to make NodeJS [client applications](https://github.com/HathorNetwork/rfcs/blob/feat/master/projects/hathor-wallet-headless/hsm-integration.md) communicate with HSM devices while there is no official SDK from Dinamo for the javascript language.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The bridge is comprised mostly of code from the [examples on Dinamo docs site](https://manual.dinamonetworks.io/c/examples.html), with added structures to facilitate communication with the NodeJS through `stdin` input and output interpretation of the `stdout` and/or `stderr`.
```
r4mmer, line 20: Since the main designed usage will be on the headless with docker, maybe we should not start a secondary process to communicate with HSM.
A known rule of thumb in docker is to have a single process per container, we are not locked from starting other processes but we should avoid where possible ([reference](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#decouple-applications))

r4mmer (commented):
Another idea would be to generate a few methods (create_xpriv, get_xpub and sign) and build this file as a shared object (.so) file so we can use this library in other projects without having to invoke the binary itself.
This makes calling the methods faster since the methods would be on-memory at all times and does not require commands to start another process to handle HSM communication. 

What do you think of this alternative?

projects/hathor-wallet-headless/hsm-c-bridge.md
tuliomir: Good points! They were added in the _Drawbacks_ and _Alternatives_ sections.

projects/hathor-wallet-headless/hsm-c-bridge.md
```diff
@@ -0,0 +1,345 @@
...
+
+  //=======================================================================
+  // Command Line Input Validation example, for alternative implementations
+  //=======================================================================
+  if (argc < 3) {
+    printf("Missing input parameters\n");
+    printf("Usage: signOnPath path {number}\n");
+    return 1;
+  }
+  nChild = atoi(argv[2]);
```
r4mmer, line 125: I know this is not the actual code that will be used, but we should somehow verify user input.

projects/hathor-wallet-headless/hsm-c-bridge.md
```diff
@@ -0,0 +1,345 @@
...
+  strncpy(authPwd.szAddr, HOST_ADDR, sizeof(authPwd.szAddr));
+  authPwd.nPort = DEFAULT_PORT;
+  strncpy(authPwd.szUserId, USER_ID, sizeof(authPwd.szUserId));
+  strncpy(authPwd.szPassword, USER_PWD, sizeof(authPwd.szPassword));
+
+  nRet = DOpenSession(&hSession, SS_USER_PWD, (BYTE *)&authPwd, sizeof(authPwd), ENCRYPTED_CONN);
+  if (nRet)
+  {
+    printf("Failure at: DOpenSession \nError code: %d\n", nRet);
+    goto clean;
```
r4mmer, line 156: I think we should avoid using goto at all.
Maybe on the actual code the session can be static global and a method would run cleanup.
Then here it would be a simple function call to cleanup state before exiting.

projects/hathor-wallet-headless/hsm-c-bridge.md
```diff
@@ -0,0 +1,345 @@
...
+### Bridge communication through TCP
+A more standardized way of communicating with the javascript applications would be through common HTTP and websocket channels.
+
+This would mean this bridge would need to become a standalone daemon and webserver application instead of a single request process, requiring a dockerized environment for itself.
+
+That would increase its maintainability at the cost of the added complexity both for code construction and process maintenance.
+
+Because of that, this solution was discarded as the first version of this client.
+
+### NodeJs to Shared Object communication
```
r4mmer, line 315: SInce this is just an alternatives sections maybe we can include that the .so source doesn't need to be in C.
We could build the bridge in rust while calling the C lib from dynamo and exposing just the rust lib.
From here we could build for any language, for instance javascript with napi-rs or python with pyo3

tuliomir:
Instead of using a C Bridge, #67 will instead use a Javascript library directly to interact with the HSM.

PR #71, opened 2023-12-12
docs: event queue connection class design

r4mmer:


projects/wallet-lib-client-for-event-queue/0002-connection-manager-design.md
```diff
@@ -0,0 +1,174 @@
...
+Will connect to the fullnode websocket and subscribe to the apropriate
+events, meaning the best block updates and the transaction events.
+
+Transaction events will be inserted into the storage and the history will be
+processed (calculating metadata, balance, etc) by the connection instance.
+Best block updates will be processed as usual.
+
+#### Events
+
+- `new-tx` and `update-tx`
```
andreabadesso, line 43: Semi-unrelated to line number, but remember to add a ping/pong mechanism

projects/wallet-lib-client-for-event-queue/0002-connection-manager-design.md
```diff
@@ -0,0 +1,174 @@
...
+of transactions from the beginning since we may have lost some events, but the
+event queue allows us to start streaming events from the moment we lost
+connection.
+This makes it so we don't have to clean the history and can just start the
+stream from where we left off.
+Although this is only applicable when we are connected to the same fullnode, so
+we need to check if the fullnode we are connected to is the same and if it isn't
+we will need to re-sync as usual.
+
+To make sure we are connected to the same fullnode we will use the peer-id, an
```
andreabadesso, line 173: We also need to assert that the `stream_id` is the same from the last sync

projects/wallet-lib-client-for-event-queue/0003-sequential-tx-processing.md
```diff
@@ -0,0 +1,137 @@
...
+
+When voiding a transaction we need to undo any impact it has on the balance of
+the wallet and the other metadata.
+This means calculating the same data as a new transaction but we should make the
+opposite changes (balances will be removed, etc).
+
+When unvoiding a transaction we can treat it as a new transaction since its
+effect on the wallet was already removed when it was voided.
+
+If the `voided_by` field is the same we can iterate on the outputs and calculate
```
andreabadesso, line 69:
```suggestion
If the `voided_by` field is empty we can iterate on the outputs and calculate
```

andreabadesso (commented):
I think it would be great to be more explicit on how the specific events from the reliable integrations feed are going to be handled, like `NEW_VERTEX_ACCEPTED` and `VERTEX_METADATA_CHANGED`

projects/wallet-lib-client-for-event-queue/0001-event-queue-connection-design.md
```diff
@@ -0,0 +1,199 @@
...
+            "group_id": null
+        }
+    ],
+    "latest_event_id": 9038
+}
+```
+
+Then we save `latest_event_id` and sync the history with the address history api.
+Once we have the wallet on the current state we can start streaming from the `latest_event_id`.
+There can be transactions arriving during this process which would mean we add them during the history sync and during the event streaming, but this issue does not affect the balance or how we process the history.
```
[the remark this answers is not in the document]
r4mmer: Balance calculation is not affected if the same transaction arrives multiple times.
This sync issue needs to be stated because we have to maintain this characteristic since the alternative would be to sync events from `id = 0` which are millions of events.

projects/wallet-lib-client-for-event-queue/0001-event-queue-connection-design.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Best block update
+
+The fullnode pubsub sends updates whenever the best chain height changes, this is so our wallets can unlock our block rewards, the event queue does not send an update like this but we receive all block transactions as events, meaning we can listen for any transaction with `version` 0 or 3 (block or merged mining block) and with the metadata `voided_by` as `null` (this is because if a block is not voided, it is on the main chain) and derive the best chain height.
+
+We will always expect the latest unvoided block to be the best chain newest block since during re-orgs (where the best chain changes) we will receive updates and the new best chain will be updated with `voided_by` as `null`.
```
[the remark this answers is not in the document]
r4mmer: We have the `current heigth` saved. If a valid and unvoided block arrives, we update the `current height`.
This works with `NEW_VERTEX_ACCEPTED` and `VERTEX_METADATA_CHANGED`.

The `voided_by` filter is just a precaution against branches on the blockchain, since the only unvoided blocks are the best chain blocks.
This also works well with re-orgs since we receive a `VERTEX_METADATA_CHANGED` to void a block but also receive one to unvoid the new best chain block, which would update the height accordingly.

projects/wallet-lib-client-for-event-queue/0001-event-queue-connection-design.md
```diff
@@ -0,0 +1,199 @@
...
+  tx_ids: string[];
+}
+```
+
+### Data conversion process
+
+To keep compatibility with the current `Connection` class we need to convert the data with the following process:
+
+1. Convert `hash` to `tx_id`
+2. Assert `nonce` is valid
```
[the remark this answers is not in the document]
r4mmer: `nonce` is required in `IHistoryTx` but it is optional on the `EventQueueTxData`.
This could be ignored, but I know the type check will complain.

projects/wallet-lib-client-for-event-queue/0001-event-queue-connection-design.md
```diff
@@ -0,0 +1,199 @@
...
+Process to convert outputs:
+
+1. Derive `token` from `token_data` and `tx.tokens`
+2. Derive `spent_by` from the output index and `tx.metadata.spent_outputs`
+3. Derive `decoded` from the script
+
+Process to convert inputs:
+
+1. Try to find the transaction with the input's `tx_id` in storage
+	1. If not found we must fetch the tx from the fullnode api
```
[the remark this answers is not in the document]
r4mmer: If `userA` spends inputs of `tx1` to send `tx2` to `userB`, we know `tx2` will be on both wallets, but `tx1` will only be on `userA`'s wallet.
The pubsub and transaction apis of the fullnode include data on all inputs since it has all transactions, but the event queue does not hidrate the transaction with more relevant input data which means we need to do this ourselves.

Basically, we need to fetch the data to do the same as these 2.

### Pubsub

When [publishing a tx](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/indexes/address_index.py#L72) it uses the `to_json_extended` method which [hidrates the inputs with the spent output data](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/transaction/base_transaction.py#L811-L818)

### Transaction API

It also uses `to_json_extended` for [list requests](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/transaction/resources/transaction.py#L277) and it does a similar [hidration](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/transaction/resources/transaction.py#L86-L91) with some extra data for single tx requests.

projects/wallet-lib-client-for-event-queue/0001-event-queue-connection-design.md
```diff
@@ -0,0 +1,199 @@
...
+2. Assign `value`, `token_data`, `script` and `token` from the spent output
+3. Derive `decoded` from the script.
```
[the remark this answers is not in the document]
r4mmer: Nice catch.

projects/wallet-lib-client-for-event-queue/0002-connection-manager-design.md
r4mmer: The document of the event queue and the implementation did not have ping/pong anywhere but from what I understand it comes with the websocket lib we use, so I'll add the ping/pong to the design

projects/wallet-lib-client-for-event-queue/0002-connection-manager-design.md
r4mmer: Nice catch.

projects/wallet-lib-client-for-event-queue/0003-sequential-tx-processing.md
r4mmer: Used `null` instead of empty but to the same effect.

andreabadesso (approved):

PR #73, opened 2024-03-28
feat: token admin methods

r4mmer:


projects/ledger-app/0001-token-admin-methods.md
```diff
@@ -0,0 +1,219 @@
...
+Since the Ledger cannot check if a token is an NFT or not, we should disregard any changes and treat both as custom tokens (which is valid if we follow the protocol).
+
+## Transaction parsing
+
+To allow admin operations we will introduce new concepts into the app that were being ignored or not treated.
+
+---
+
+The `token_data` of an output is a byte with the information:
+- Weather the token is a custom token or HTR
```
msbrogli, line 38: Typo: whether

projects/ledger-app/0001-token-admin-methods.md
```diff
@@ -0,0 +1,219 @@
...
+	- Other authorities will throw.
+	- We will not allow joined authority, for instance `mint+melt authority` (i.e. `value = 3`)
+- Approve step
+- Reject step
+
+Obs: With authority outputs we already have enough to enable mint, melt, delegate and destroy operations.
+
+---
+
+Currently we only allow P2PKH scripts (and addresses) but NFTs have a data output.
```
msbrogli, line 62: What about P2SH scripts?

projects/ledger-app/0001-token-admin-methods.md
```diff
@@ -0,0 +1,219 @@
...
+3. There will be an output with the created token, the token uid will not be on the tokens array since it does not exist yet
+	1. Token uid is the transaction id of the created transaction, which is created during mining.
+
+To deal with these we will create a new command to load the token data of the token being created.
+We will use the existing `SEND_TOKEN_DATA` for instruction code `0x08` and use the `P1` field to indicate which data is being sent.
+
+| Description                                 | P1   | P2   | CData                                                                                                                                                                   |
+| ------------------------------------------- | ---- | ---- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Send token data<br>for signing transactions | 0x00 | 0x00 | `version (1)` \|\|<br> `uid (32)` \|\|<br>`symbol_len (1)` \|\|<br>`symbol (symbol_len)` \|\|<br>`name_len (1)` \|\|<br>`name (name_len)` \|\|<br>`signature (32)`      |
```
msbrogli, line 104: This command seems to be very specific. Why not send the transaction bytes with a desired type and let the Ledger app parse and handle it? This way the Ledger app would support more advanced use cases without any change.

projects/ledger-app/0001-token-admin-methods.md
```diff
@@ -0,0 +1,219 @@
...
+> [!NOTE] Version field has been split into `signalBits` and `version`
+> The version field of the transaction was 2 bytes (first byte was reserved for later use) and since Ledger app version 1.1.0 the version field has become a 1 byte field and the first byte became the `signalBits`
+
+Version byte needs to be refactored to a 1 byte field and support for `signalBits` will not be added in this design so we will only allow `0x00` as `signalBits` for now.
+
+## Admin operations
+
+### Mint, Melt, Delegate and Destroy
+
+Sending a mint, melt, delegate or destroy transactions should use the same protocol as sending normal transactions.
```
msbrogli, line 83: How are we going to present which is the token id that is being minted/melted/delegated/destroyed? I feel that we should present the Token Name, Token Symbol and Token UID.

I also think that the wallet should send the TokenCreationTransaction to the Ledger app, so it doesn't have to trust metadata. Then, the Ledger app will calculate the hash, confirm it is the same as the token_id in the new transaction, and, finally, get the token name and token symbol from the TokenCreationTransaction.

msbrogli (changes requested):

projects/ledger-app/0001-token-admin-methods.md
r4mmer: Under "Future possibilities" is discussed the difficulties with P2SH, while this design may not comtemplate P2SH support we may add partial or full support in the future.
Partial support would entail showing P2SH addresses on the transaction confirmation.
Full support would entail address generation (basically a P2SH mode on the app) and recognizing P2SH change outputs.

projects/ledger-app/0001-token-admin-methods.md
r4mmer: > How are we going to present which is the token id that is being minted/melted/delegated/destroyed? I feel that we should present the Token Name, Token Symbol and Token UID.

Since the Ledger does not have the transaction from the inputs we are somewhat limited here.
Having all spent transactions sent to Ledger is not feasible either since it may not have enough memory for this (nanoS has about 4.5Kb of RAM).
So a mint operation that does not create a new mint authority is indistinguishable (from the outputs only) from a transaction that is sending some tokens, we may add some information to make the UX nicer, but the ledger may not be able to check that mint/melt/destroy operations are being done (delegate operations always have an authority output).
An example of this is that if we do not send authority outputs we can already mint and melt tokens and destroy token authorities with the Ledger device.

> I also think that the wallet should send the TokenCreationTransaction to the Ledger app, so it doesn't have to trust metadata. Then, the Ledger app will calculate the hash, confirm it is the same as the token_id in the new transaction, and, finally, get the token name and token symbol from the TokenCreationTransaction.

Currently the user has to manually confirm any tokens that the he wants to send in transactions with the Ledger, this process is called token signing and must be done before we attempt to sign a transaction with the token (the app will block a transaction if a token in the token array was not signed by the user).
We could add the "TokenCreationTransaction hash matching the UID" check to this token signing to make it even safer, but I think we should avoid blindly trusting a token to be sent using the Ledger just because it matches the TokenCreationTransaction.
Also, I think this would merit a design of its own.

projects/ledger-app/0001-token-admin-methods.md
```diff
@@ -0,0 +1,219 @@
...
+	- We will not allow joined authority, for instance `mint+melt authority` (i.e. `value = 3`)
+- Approve step
+- Reject step
+
+Obs: With authority outputs we already have enough to enable mint, melt, delegate and destroy operations.
+
+---
+
+Currently we only allow P2PKH scripts (and addresses) but NFTs have a data output.
+Data outputs are allowed on the protocol but we will only allow on NFT creation scripts because any other is just an HTR burn.
```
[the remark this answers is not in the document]
r4mmer: Do we want to allow any data output on any transaction?
If so I can change it.

I just think that the app should be restrictive on what we want to allow, that's why I added this but as stated in the design there is no issue either on the app or the protocol to add support for data outputs.

If we want to add support for data outputs we just need to confirm which encoding we want to show the user.
The NFT data output would use ascii (subset of UTF-8) so the user con confirm the NFT url but if its not an NFT creation I think we should use hex encoding (to allow any byte string).
What do you think?

Obs: data scripts are defined as scripts that could not be parsed, meaning they are not P2SH or P2PKH. We currently do not check for P2SH so to add support for all data outputs we would need to add P2SH parsing, unless we accept a P2SH script showing as a data script until we add support for P2SH.

projects/ledger-app/0001-token-admin-methods.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Mint, Melt, Delegate and Destroy
+
+Sending a mint, melt, delegate or destroy transactions should use the same protocol as sending normal transactions.
+Mint and Melt involve destroying/creating custom token outputs and since the Ledger app does not have access to the input data it can only see the outputs, which means that a mint operation is a transaction with custom tokens outputs (possibly HTR outputs for change) and authority outputs, same for melt operations.
+
+Delegate is simpler since its actually just sending an authority output.
+
+Destroy operations usually do not even have authority outputs, so it actually is already supported.
+Granted that the token array will have the token uid of the authority being destroyed so the user still has to sign the token data even with no outputs of the token.
```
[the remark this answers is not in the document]
r4mmer: We cannot know this since the app does not have any information on the inputs.

projects/ledger-app/0001-token-admin-methods.md
r4mmer: Changed

projects/ledger-app/0001-token-admin-methods.md
```diff
@@ -0,0 +1,219 @@
...
+> The version field of the transaction was 2 bytes (first byte was reserved for later use) and since Ledger app version 1.1.0 the version field has become a 1 byte field and the first byte became the `signalBits`
+
+Version byte needs to be refactored to a 1 byte field and support for `signalBits` will not be added in this design so we will only allow `0x00` as `signalBits` for now.
+
+## Admin operations
+
+### Mint, Melt, Delegate and Destroy
+
+Sending a mint, melt, delegate or destroy transactions should use the same protocol as sending normal transactions.
+Mint and Melt involve destroying/creating custom token outputs and since the Ledger app does not have access to the input data it can only see the outputs, which means that a mint operation is a transaction with custom tokens outputs (possibly HTR outputs for change) and authority outputs, same for melt operations.
```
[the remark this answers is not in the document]
r4mmer: Changed

projects/ledger-app/0001-token-admin-methods.md
```diff
@@ -0,0 +1,219 @@
...
+[motivation]: #motivation
+
+Today Hathor's Ledger app only allow sending transactions with custom tokens, so users and use-cases can hold and send custom tokens but to create and manage token supply it is required to use a software wallet.
+This design is to allow users who use Ledger, a hardware cold wallet, to fully manage custom tokens without using any software wallet.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+> [!WARNING] Request for comment
+> This whole paragraph below can change if we really want to, we just need to add an `isNFT` boolean in the token data for the user to confirm, so in all transactions the Ledger will be informed that the token is an NFT with the user express consent during token signing
```
msbrogli, line 22: I would use a `number of decimals` parameter (instead of `isNFT`). In this case, NFTs would use `decimals=0`.

projects/ledger-app/0001-token-admin-methods.md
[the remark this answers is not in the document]
pedroferreira1: So which informations are we going to request the user to confirm, if the tx is a simple authority destroy? You usually confirm only the outputs? 

How is the flow going to be?

projects/ledger-app/0001-token-admin-methods.md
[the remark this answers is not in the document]
r4mmer: > We have a way to parse the output script to identify its type, we could use it, can't we?

But we dont have a way to identify the script, we would need to implement one.
The idea would be to check if its P2PKH, then check P2SH and if none of the above match, it is a data output.
The "check if its P2SH" method does not exist currently.

> my main concern here is not adding this support but that the code to handle the NFT data output is not too specific

I think making it very specific is what gives us more flexibility in the future.
The case of the design is only for "Create token transactions" and only for the first data output on them.
In the future we can add support for data outputs in other places since it won't interfere with this specific case.

projects/ledger-app/0001-token-admin-methods.md
r4mmer: Adding this new parameter will be addressed in another design since it requires some discussion on UI/UX (e.g. if we should add scrolling when using a lot of decimals).

projects/ledger-app/0001-token-admin-methods.md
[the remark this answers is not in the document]
pedroferreira1: Does it make sense to have a screen with "This transaction has no outputs" for the user to confirm?

pedroferreira1 (dismissed):

pedroferreira1 (approved):

PR #74, opened 2024-04-10
feat(feature-activation): add miner update doc

glevco:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/feat/feature-activation/miner-update/projects/feature-activation/0005-miner-update-request.md)

### DIFF
--- a/projects/feature-activation/0005-miner-update-request.md
+++ b/projects/feature-activation/0005-miner-update-request.md
@@ -1,0 +1,162 @@
+# Hathor Integration Update Request
+
+## Introduction
+
+We’re currently in the testing phase of a project called Feature Activation, which is a mechanism used to activate new consensus-agreed features on Hathor Network. In the future, it will be used to release important features such as Nano Contracts and increasing the maximum Merkle Path length, which will allow miners to find more valid merge mined blocks. Here’s a [blog post](https://blog.hathor.network/hathors-feature-activation-process-miners-lead-the-way-d7f43b12978d) about it.
+
+In order for that to work, we need miners to propagate a support signal for features through their mined blocks. This is simply a repurposed byte in the block data. We’ll soon have a complete guide on how to interact with feature signals, but everything should work by default and transparently. However, at this time, we kindly request your cooperation.
+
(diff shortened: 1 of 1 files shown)


projects/feature-activation/0005-miner-update-request.md
```diff
@@ -0,0 +1,61 @@
...
+
+We of course are available for any questions or help that you might need. Thank you in advance!
+
+## Updating the Integration
+
+In order to support this new feature, a simple change must be made.
+
+### 1. Identifying where the Hathor block is generated
+
+Your current integration most certainly reads our Block template from one of our HTTP or WebSocket endpoints. The fields provided in those endpoints are the following:
```
glevco, line 19: Can we be more specific here, by mentioning the specific API they use?

projects/feature-activation/0005-miner-update-request.md
```diff
@@ -0,0 +1,61 @@
...
+parents_any: list[bytes]
+height: int
+score: float
+```
+
+Identify where your integration code reads this data and constructs the block.
+
+### 2. Adding `signal_bits` to the block data
+
+Currently, you integration ignores the `signal_bits` field, which didn't exist before. You likely set the first byte of the block data as `0x00`, or just use the `version` field to set the first two block bytes. You must change this so the `signal_bits` field is the first block byte, and the `version` field is the second block byte.
```
jansegre, line 39: you integration -> your integration

Also worth mentioning that every previous version of the full node will ignore the value of the first byte, so using the `signal_bits` will not affect the propagation of blocks to peers that have not updated to the latest version.

jansegre (dismissed):

projects/feature-activation/0005-miner-update-request.md
glevco: @jansegre ?

msbrogli: They could also be using the ``/v1a/get_block_template` HTTP API. So we might mention it in the guide.

msbrogli (approved):

projects/feature-activation/0005-miner-update-request.md
glevco: Added in https://github.com/HathorNetwork/rfcs/pull/74/commits/9cb72cc16011c3ff7ccdb4ab91a2b6518dbcee52

jansegre: I remember there being 3 different mining APIs, an older HTTP, a newer HTTP and a WebSocket, the newer HTTP and WebSocket should have the same fields (for the most part). I'll take a look to confirm this.

jansegre (approved):

PR #77, opened 2024-05-09
design: improved pin security mode

r4mmer:
[rendered](https://github.com/HathorNetwork/rfcs/blob/docs/improved-pin-security-mode/projects/hathor-wallet-mobile/security/0001-improved-pin-security-mode.md)

### DIFF
--- a/projects/hathor-wallet-mobile/security/0001-improved-pin-security-mode.md
+++ b/projects/hathor-wallet-mobile/security/0001-improved-pin-security-mode.md
@@ -1,0 +1,66 @@
+# Summary
+[summary]: #summary
+
+A new way to use biometrics on the mobile wallet, using a longer pin for security, this does not allow a fallback to the configured pin in case of too many failures.
+
+# Motivation
+[motivation]: #motivation
+
+When using biometrics the pin is saved on the phone system, meaning the storage data is still encrypted with the 6 digit pin, using a longer and more random pin makes the storage data harder to crack.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When enabling the biometrics on the mobile wallet we just set a flag and the pin is always saved on the device keychain, this makes it so enabling and disabling biometrics does not require the user to input the pin.
+This approach makes it easier to set up biometrics and allows the usual pin screen to be used as a fallback.
+
+The new configuration would use the same configuration for biometrics but use a longer and unguessable password (random 32 bytes for instance), it will trust the system prompt completely, so being locked out due to too many attemps will lock the user out of the wallet.
+
+## Access control
+
+The dependency used to manage keychain password is [react-native-keychain](https://www.npmjs.com/package/react-native-keychain) and we use the [setGenericPassword](https://github.com/oblador/react-native-keychain?tab=readme-ov-file#setgenericpasswordusername-password--accesscontrol-accessible-accessgroup-service-securitylevel-) and [getGenericPassword](https://github.com/oblador/react-native-keychain?tab=readme-ov-file#getgenericpassword-authenticationprompt-service-accesscontrol-) to save and retrieve the pin from the keychain.
+Since the new security mode and the current biometrics mode cannot work at the same time we will only use 1 password saved at a time on the keychain (instead of having both pins saved).
+
+The [access control](https://github.com/oblador/react-native-keychain?tab=readme-ov-file#keychainaccess_control-enum) and [accessible](https://github.com/oblador/react-native-keychain?tab=readme-ov-file#keychainaccessible-enum) configurations are used with the `setGenericPassword` to manage how this password can be used and by whom.
+
+This new password should be saved with:
+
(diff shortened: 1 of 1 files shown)


projects/hathor-wallet-mobile/security/0001-improved-pin-security-mode.md
```diff
@@ -0,0 +1,66 @@
+# Summary
+[summary]: #summary
+
+A new way to use biometrics on the mobile wallet, using a longer pin for security, this does not allow a fallback to the configured pin in case of too many failures.
+
+# Motivation
+[motivation]: #motivation
+
+When using biometrics the pin is saved on the phone system, meaning the storage data is still encrypted with the 6 digit pin, using a longer and more random pin makes the storage data harder to crack.
```
alexruzenhack, line 9: This makes me wonder if we should also review the security mode for the pin itself. But I believe it is for "future possibility".

alexruzenhack (approved):

alexruzenhack (dismissed):

projects/hathor-wallet-mobile/security/0001-improved-pin-security-mode.md
```diff
@@ -0,0 +1,66 @@
...
+Since the new security mode and the current biometrics mode cannot work at the same time we will only use 1 password saved at a time on the keychain (instead of having both pins saved).
+
+The [access control](https://github.com/oblador/react-native-keychain?tab=readme-ov-file#keychainaccess_control-enum) and [accessible](https://github.com/oblador/react-native-keychain?tab=readme-ov-file#keychainaccessible-enum) configurations are used with the `setGenericPassword` to manage how this password can be used and by whom.
+
+This new password should be saved with:
+
+- Access control **`BIOMETRY_ANY_OR_DEVICE_PASSCODE`**: Constraint to access an item with Touch ID for any enrolled fingers or passcode.
+	- This way either biometrics or system pin/pattern will unlock the wallet
```
pedroferreira1, line 29: I'm wondering if we should allow this. This makes anyone that knows the phone passcode to send a transaction using the wallet, I'm not sure if it makes sense.

projects/hathor-wallet-mobile/security/0001-improved-pin-security-mode.md
```diff
@@ -0,0 +1,66 @@
...
+- Accessible **`WHEN_UNLOCKED_THIS_DEVICE_ONLY`**: The data in the keychain item can be accessed only while the device is unlocked by the user. Items with this attribute do not migrate to a new device.
+
+## Settings
+
+The new configuration should not replace the current one, it will be presented to the user as an option called "Enable system authentication" on the settings page, once clicked the enabling process will start on a new screen.
+
+### Enabling process
+
+When trying to enable system authentication the user will be required to confirm that this is what he wants, this screen also has a more detailed explanation of what will be enabled along with a warning that the wallet pin unlock will be disabled.
```
pedroferreira1, line 38: Does it make sense to have both ways? When I was reading, I thought we would allow only for the user to have the long password with biometrics, or the 6 digit pin without biometrics.

I think it becomes really confusing having this so customizable.

r4mmer: Personally I think that this is a good default for security, but not being able to fallback to the pin may cause issues so i included in the design a way to work as a new feature along with the old biometry mode.

If we decide to only have 1 biometry mode available we could use the feature flag that enables this mode to also disable the old mode, so users will only have 1 biometry mode available for them and gradually we could disable the old biometry mode.

projects/hathor-wallet-mobile/security/0001-improved-pin-security-mode.md
r4mmer: We could change to `BIOMETRY_ANY` or even have the user decide if he wants biometry or device passcode.

r4mmer: Another point that I was thinking about this is that today we have the access data saved on storage and we check the pin against the main key which allows us to start but the xpub is saved unencrypted so we could start (scan addresses) without the pin.

We could just encrypt the entire access data with the biometry password so the biometry would allow the wallet to start and the user would have to use the 6 digit pin to send transactions, much like some banking apps which have biometry or a app/web code and if you want to make a transaction you need another passcode.

pedroferreira1: I like this approach, but I think we should discuss this UX change in another design

projects/hathor-wallet-mobile/security/0001-improved-pin-security-mode.md
pedroferreira1: I like that you added it to the future possibilities.

pedroferreira1 (approved):

alexruzenhack (approved):

PR #82, opened 2024-07-31
feat: including Dozer proposal for token economics

tothster:
This is a proposal for a review in Hathor token economics for the custom tokens create process.


obiyankenobi:
Another unresolved question: will the fees incur when the token is deposited to a nano contract?

hBulldozer:
I would include the fees for tokens in the new model for the Nano contract Actions. Eg. token transfers between users<> contracts, and contracts<>contracts.

PR #93, opened 2025-01-07
docs: added nano contracts support for the wallet-service

andreabadesso:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/3d0ebcc85d97fc89fbebcd9db487facc39b73784/projects/wallet-service/nano-contracts/0001-nano-contracts-support.md)

### DIFF
--- a/projects/wallet-service/nano-contracts/0001-nano-contracts-support.md
+++ b/projects/wallet-service/nano-contracts/0001-nano-contracts-support.md
@@ -1,0 +1,165 @@
+- Feature Name: nano-contracts-support
+- Start Date: 2025-01-02
+- Author: @andreabadesso
+
+# Summary
+
+This document describes the integration of nano-contracts support into the wallet-service, enabling tracking and indexing of nano-contract transactions, including caller information and providing proxy endpoints for fullnode interaction.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


pedroferreira1 (approved):
For me this is approved. I just have two considerations:

1. We will need to handle on chain blueprint transactions the same way as the nano contracts, so you can also add them in this RFC. They have signature and pubkey as the nano contract and I don't thin we need to add the code, just the transaction, for the same reason as the nano contract (to keep the history and tx count consistent).
2. While discussing the new full node APIs for the explorer in relation to nano contracts, implementing certain features—such as sorting by transaction count or the timestamp of the latest nano transaction—proved to be quite challenging. We will need to discuss in the future ways to store blueprint informations (even built-in), so we can to this API in the wallet service / data service / elastic search somehow.

tuliomir (approved):
This design looks good to me and allows the wallet service to interact with the nano contracts improvements on the network.

I tried thinking of possibilities, however, to optimize the amount of accesses to the fullnode APIs, as I understand this is one of the main objectives of the Wallet Service.

One idea that occurred to me is to cache responses from calls to specific nano contracts for the duration of the current block. That, however, would be a simple optimization that would only benefit nanocontracts with high interest, and probably wouldn't be useful at this moment of the network.

So, for now I only leave this as an improvement idea as reference for future brainstorms and features.

PR #94, opened 2025-02-24
docs: adds transaction fee minting rfc

raul-oliveira:


text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+  - Set **minimum transaction fees** to prevent microtransactions from overloading the system.
+
+---
+
+### 3. Complexity in Multi-Token Transactions
+
+**Issue:** Transactions involving multiple tokens introduce a more complex fee structure, requiring additional computation to determine the appropriate charge.
+
+**Impact:**
+  - Increased processing overhead for nodes.
```
msbrogli, line 169: I'd say it would increase processing overhead for wallets (not nodes).

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+**Potential Mitigation:**
+  - Improve **wallet UX** to clearly display estimated fees before confirming transactions.
+  - Optimize **fee calculation logic** to minimize the processing impact.
+
+---
+
+### 4. Miner Incentive Model May Need Adjustments
+
+**Issue:** Since miners are currently incentivized through TX mining and HTR block rewards, a shift to fee-based transactions could impact their revenue model.
```
msbrogli, line 180: Tx mining is not part of the miners incentive scheme. Additionally, I can't see how the current proposal would impact miners since all previous mechanisms would still be in place.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+  - Allow **fee-burning mechanisms** to ensure that fees remain valuable to miners.
+
+---
+
+### 5. Interactions with Smart Contracts and Additional Execution Costs
+
+**Issue:** Nano contracts and other execution-based transactions introduce **variable computational costs** that are not reflected in a simple per-transaction fee model.
+
+**Impact:**
+  - Fees may be **too low to cover execution costs**, leading to **free rider issues**.
```
msbrogli, line 197: What are free rider issues?

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+To address the spam problem, some networks have tested models with proof of work (PoW) embedded in transactions. This mechanism requires users to perform a small computational task before sending transactions, making mass spam transactions more difficult. Another approach is the use of dynamically adjusted microfees to maintain network balance.
+
+This model offers several advantages. First, it simplifies the user experience by providing a predictable fee without major fluctuations. Additionally, it enables a filtering mechanism to prevent excessively large or complex transactions, protecting the network from spam. Finally, its structure promotes sustainability, ensuring controlled blockchain growth.
+
+### General Fee calculation rule
+
+1. HTR is always used as the fee unit, regardless of the tokens involved in the transaction.
+2. If only fee-based tokens are used, a single fee in HTR will be charged, regardless of the number of tokens.
```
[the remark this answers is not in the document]
raul-oliveira: Inside the example scenarios we have a description of this use case. In any case I'll add to the document a formula to explain the fee calculation

#### Token A x Token B x Token C (only 1 fee, in HTR)

Description: Multi-token transfer involving three fee-based tokens.

Fee Calculation:
- Only one HTR fee is charged, regardless of the number of tokens.
```
- Base Fee: 0.001 HTR
- Complexity Fee (extra tokens): 0.0005 HTR per additional token.
- Total Fee: 0.002 HTR (0.001 + 0.0005 + 0.0005)

File: text/0045-transaction_fee_minting.md [hunk 5]
```diff
@@ -0,0 +1,232 @@
...
+# Motivation
+[motivation]: #motivation
+
+Dozer suggested an alternative to the HTR deposit requirement when minting tokens. The idea is to remove the 1% deposit requirement and, instead, implement a transaction fee for every transfer of the minted tokens. [RFC](https://github.com/Dozer-Protocol/hathor-rfcs/blob/new-token-economics/projects/new-token-economics/token-economics.md)
+
+This change would reduce the upfront cost of minting tokens, making it more accessible to users who may not have sufficient HTR at the time of minting. The network would still benefit from a fee mechanism that contributes to miners’ incentives and overall network security.
+
+An explicit use case is meme coins creation.
+
+# Guide-level explanation
```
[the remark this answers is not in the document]
raul-oliveira: IMO we should control inflation, and reward miners. In any case, I don't have enough experience with blockchains to decide it.
It's a question to be addressed by the business. The business should decide the kind of the drawbacks it would take.

text/0045-transaction_fee_minting.md
raul-oliveira: fixed 👍🏼

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,175 @@
...
+[motivation]: #motivation
+
+Dozer suggested an alternative to the HTR deposit requirement when minting tokens. The idea is to create a new type of custom tokens where tokens would be minted for free (i.e., no deposits) and fess will be charged for transactions. [RFC](https://github.com/Dozer-Protocol/hathor-rfcs/blob/new-token-economics/projects/new-token-economics/token-economics.md)
+
+This change would reduce the upfront cost of minting tokens, making it more accessible to users who may not have sufficient HTR at the time of minting. The network would still benefit from a fee mechanism that contributes to miners’ incentives and overall network security.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When minting tokens, creators can choose between two transaction models: Deposit-Based and Fee-Based. Each model offers different benefits, depending on how the tokens will be used. For handle it we will use the same definitions of the [Token Creation Transaction](https://github.com/HathorNetwork/rfcs/blob/master/text/0004-tokens.md), possibly adding a variation of it using a new version.
```
raul-oliveira, line 23: I'm still validating which version (tx version or token_info_version) to change with a POC

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+**Potential Mitigation:**
+  - Introduce a **burn mechanism** where a percentage of the transaction fee is permanently removed from circulation.
+  - Implement an **optional melt requirement** where token issuers must periodically remove supply from circulation.
+
+---
+
+### 2. Potential for Low-Cost Network Congestion
+
+**Issue:** If fees are too low, bad actors could create a high volume of low-value transactions, leading to congestion.
```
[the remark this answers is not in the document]
raul-oliveira: I removed this piece, since the PoW solve the issue. Tks for the explanation

text/0045-transaction_fee_minting.md
raul-oliveira: When some users benefit from network services without contributing a fair share to the costs, potentially straining the system’s sustainability. 

I removed the bullet since is something that should be addressed by nano, not by this proposal.

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
raul-oliveira: I'll spend some time understanding better the PoW to answer this one, back to you soon as I get the answer

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,167 @@
...
+# Summary
+[summary]: #summary
+Currently, when minting X tokens, a [deposit of P% (1%) of X in HTR is required](./0011-token-deposit.md). This deposit can later be withdrawn when the tokens are melted. 
+
+This proposal suggests an alternative mechanism where, instead of requiring an upfront deposit of HTR when minting new tokens, a transaction fee is charged on each transfer of the newly minted tokens.
+
+# Motivation
+[motivation]: #motivation
+
+Dozer suggested an alternative to the HTR deposit requirement when minting tokens. The idea is to create a new type of custom tokens where tokens would be minted for free (i.e., no deposits) and fess will be charged for transactions. [RFC](https://github.com/Dozer-Protocol/hathor-rfcs/blob/new-token-economics/projects/new-token-economics/token-economics.md)
```
glevco, line 16: Typo: fess

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Deposit-Based Model (as is)
+
+Currently, creators must [deposit a percentage](https://github.com/HathorNetwork/rfcs/blob/master/text/0011-token-deposit.md) (P%) of the minted tokens’ value in HTR. For example, if P% is set at 1% and a user mints 100 tokens, they must deposit 1 HTR. This deposit acts as a reserve and is fully refundable if the tokens are later melted.
+
+## Fee-Based Model
+
+Adding the fee-based model the platform won't require an upfront deposit. Instead, each transfer of the minted tokens incurs a transaction fee, which is deducted automatically. The exact fee rate will be defined at a later stage.
+
+By selecting the appropriate model, token creators can optimize their minting strategy based on their specific needs and usage scenarios. A key use case is the creation of memecoins.
```
glevco, line 33: Why memecoins? I think it's because they are usually created with a lot of tokens, right?

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,167 @@
...
+
+### How to pproach
```
glevco, line 35: Typo: pproach

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,167 @@
...
+
+An alternative to bitcoin and etherium approchaes for blockchain fees is a hybrid model that combines fees based in the transaction size, with priority tip payments.
```
glevco, line 37: Typo: etherium

glevco, line 37: What's priority tip?

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,167 @@
...
+1. HTR is always used as the fee unit, regardless of the tokens involved in the transaction.
+2. Even if fee-based tokens are used alone, or with a deposit-based token, a single fee in HTR will be charged, regarding the transaction size.
+3. Interacting with nano contracts may introduce an additional fee depending on the execution complexity.
+
+Below is a pseudo code that explain how fee should be calculated, using a fee rate per byte:
+
+*(Note: Excluded "hash" size from calculation, hence 114 reduced to 84 bytes.)*
+
+```python
+def calculate_tx_fee(tx, fee_rate_per_byte):
```
glevco, line 57: You don't need all this. Instead, you can simply serialize the tx to bytes and get its total size. The DAA does exactly that to calculate the tx weight, for reference: https://github.com/HathorNetwork/hathor-core/blob/80c5ae344403795d08ef3ec07a37b1c5f120af23/hathor/daa.py#L198

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,167 @@
...
+|----------------------|-----------------------------|----------|
+| Fixed-size fields    |                             | **84** bytes |
+| Tokens               | 1 token × 32 bytes          | **32** bytes |
+| Inputs               | 2 × (35 + 108)              | **286** bytes |
+| Outputs              | 2 × (7 + 25)                | **64** bytes |
+| **Total Tx Size**    | 84 + 32 + 286 + 64          | **466** bytes |
+
+## Wallet UI
+
+We should provide to the users the option to select between the two models (deposit and fee) when minting.
```
glevco, line 106: Should we also mention that explorer UIs need to be updated?

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,167 @@
...
+
+### How to pproach
+
+An alternative to bitcoin and etherium approchaes for blockchain fees is a hybrid model that combines fees based in the transaction size, with priority tip payments.
+
+This model offers several advantages. First, it simplifies the user experience by providing a predictable fee without major fluctuations. Additionally, it enables a filtering mechanism to prevent excessively large or complex transactions, protecting the network from spam. Finally, its structure promotes sustainability, ensuring controlled blockchain growth.
+
+### Fee calculation
+
+Hathor’s fees will be calculated similarly to Bitcoin, depending only on transaction size (in bytes) and a network-defined fee rate (in HTR per byte).
```
obiyankenobi, line 43: Why are we going for this model? The original proposal suggests a flat fee.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+1. HTR is always used as the fee unit, regardless of the tokens involved in the transaction.
+2. If only fee-based tokens are used, a single fee in HTR will be charged, regardless of the number of tokens.
+3. If a transaction includes both deposit-based and fee-based tokens, only the fee-based tokens will incur a fee.
+4. Interacting with nano contracts may introduce an additional fee depending on the execution complexity.
+
+### Example Scenarios
+
+####  Token A x Token A (fee in HTR)
+
+Description: Transfer of the same token (A) between two parties.
```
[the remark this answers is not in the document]
raul-oliveira: I changed it to swap in order to clarify

text/0045-transaction_fee_minting.md
raul-oliveira: Yes, you got it. This is one popular use case, but the user can use for any other scenario that might appear.

text/0045-transaction_fee_minting.md
raul-oliveira: done

text/0045-transaction_fee_minting.md
raul-oliveira: done

text/0045-transaction_fee_minting.md
raul-oliveira: In Ethereum, a priority tip (also called a priority fee) is an optional payment added to transactions after the London Upgrade (EIP-1559) to incentivize miners (or validators) to include your transaction in the next block promptly.

text/0045-transaction_fee_minting.md
raul-oliveira: Since Hathor follows the line to be easy to use, make sense for me to have an ui to reflect our features.

text/0045-transaction_fee_minting.md
raul-oliveira: Thank you! It is very helpful. I was wondering just to explain that we have a fixed size and a variable size. To compose a base-fee and a dynamic fee.

Should I update this piece anyway?

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
raul-oliveira: Done

text/0045-transaction_fee_minting.md
raul-oliveira: How do you see tokens which value nothing being used to pay the fee in a transaction?

Example:
- Token A is minted without a deposit, and I can choose the amount.It means token A values nothing. 

Does make sense to enable the charge of custom tokens?

hBulldozer:
It's always wonderful to see our suggestions being put into action. Thank you for listening to us!

We're posting this to provide context on our thinking behind a possible feature allowing projects to charge fees using custom tokens created with a 1% HTR lock requirement. 

We thought of a way to give utility to a specific token, just like HTR is used for the overall network. A project could create tokenA that they control the supply and distribution of and can sell for a fixed price for example. And another tokenB that would be the transaction token for their service, so for users to use the service related to tokenB, they would need to have tokenA.

This tokenA would be like a credit/gas token for this protocol. Since it would have been created with HTR, 100x more of it would have been burned, effectively giving utility to the locked HTR while also benefiting the ecosystem of the project that distributes their project's tokenA, credit/gas token.

Thinking about it more now, I see that it seems complex and is a use case that may have little adoption, but when we wrote the RFC, we thought it wouldn't be complex to implement and could be interesting.

In some projects, people modify the ERC20 to include a transaction fee in the token itself that can be burned or sent to the treasury. However, we understood that this wouldn't add utility to HTR, which is why we suggested using this "gas” token.

We agree with Yan and Trond that following Solana's low flat fee approach is a better strategy for the ecosystem. While it may not drive significant demand for HTR in individual transactions, it will become meaningful with the expected volume of small transactions from users/agents/services currently excluded from Ethereum and Bitcoin due to their high, unpredictable fees.​​​​​​​​​​​​​​​​

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+# Summary
+[summary]: #summary
+Currently, when minting X tokens, a [deposit of P% (1%) of X in HTR is required](./0011-token-deposit.md). This deposit can later be withdrawn when the tokens are melted. 
+
+This proposal suggests an alternative mechanism where, instead of requiring an upfront deposit of HTR when minting new tokens, a transaction fee is charged on each transfer of the newly minted tokens.
+
+# Motivation
+[motivation]: #motivation
+
+Dozer suggested an alternative to the HTR deposit requirement when minting tokens. The idea is to remove the 1% deposit requirement and, instead, implement a transaction fee for every transfer of the minted tokens. [RFC](https://github.com/Dozer-Protocol/hathor-rfcs/blob/new-token-economics/projects/new-token-economics/token-economics.md)
```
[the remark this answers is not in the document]
msbrogli: Thanks!

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+An explicit use case is meme coins creation.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When minting tokens, creators can choose between two transaction models: Deposit-Based and Fee-Based. Each model offers different benefits, depending on how the tokens will be used.
+
+## Deposit-Based Model
+
+In this [model](https://github.com/HathorNetwork/rfcs/blob/master/text/0011-token-deposit.md), creators must deposit a percentage (P%) of the minted tokens’ value in HTR. For example, if P% is set at 1% and a user mints 100 tokens, they must deposit 1 HTR. This deposit acts as a reserve and is fully refundable if the tokens are later melted.
```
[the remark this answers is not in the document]
msbrogli: Ok.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Fee-Based Model
+
+The fee-based model does not require an upfront deposit. Instead, each transfer of the minted tokens incurs a transaction fee, which is deducted automatically. The exact fee rate will be defined at a later stage.
```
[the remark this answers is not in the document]
msbrogli: Ok.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+By selecting the appropriate model, token creators can optimize their minting strategy based on their specific needs and usage scenarios. A key use case is the creation of memecoins.
+
+### How Bitcoin deals with fee
```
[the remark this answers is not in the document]
msbrogli: Thanks!

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+Bitcoin handles transaction fees through a system based on competition for limited block space. Each transaction has a weight measured in vBytes, and users set their fees by offering satoshis per vByte, which influences the confirmation speed. During periods of high demand, fees increase due to competition, while in lower traffic periods, they decrease. 
+
+Wallets and services use fee estimators to suggest optimal fees based on the mempool. The protocol does not enforce a minimum fee, but nodes can set a "relay fee" to prevent spam from very low-value transactions.
+
+### How Etherium deals with fee
```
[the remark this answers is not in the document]
msbrogli: Thanks!

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+The Base Fee is burned, reducing the Ether supply, while the Priority Tip goes to validators. This model improves fee predictability, making costs more stable and transparent for users.
+
+### An hybrid approach
+
+A third option for blockchain fees is a hybrid model that combines minimum fees with payments based on actual resource usage, along with policies to prevent spam on the network. This model aims to balance user simplicity, protection against spam attacks, and long-term sustainability.
+
+The adaptive minimum fee consists of a fixed base fee that all users pay per transaction, ensuring a minimum cost for network usage and preventing abuse. Additionally, there is a variable component that adjusts the fee based on resource consumption, such as transaction size or script complexity.
+
+To address the spam problem, some networks have tested models with proof of work (PoW) embedded in transactions. This mechanism requires users to perform a small computational task before sending transactions, making mass spam transactions more difficult. Another approach is the use of dynamically adjusted microfees to maintain network balance.
```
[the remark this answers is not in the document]
msbrogli: It seems you removed these paragraphs. Is that correct?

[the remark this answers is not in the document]
raul-oliveira: Yes, I moved it to Prior art

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+Wallets and services use fee estimators to suggest optimal fees based on the mempool. The protocol does not enforce a minimum fee, but nodes can set a "relay fee" to prevent spam from very low-value transactions.
+
+### How Etherium deals with fee
+
+Ethereum handles transaction fees through the concept of Gas, where each operation consumes a specific amount of computational resources. Before EIP-1559, users set the Gas Price, and those who paid more had priority. With EIP-1559, a dynamically adjusted Base Fee was introduced based on demand, along with an optional Priority Tip to incentivize validators. 
+
+The Base Fee is burned, reducing the Ether supply, while the Priority Tip goes to validators. This model improves fee predictability, making costs more stable and transparent for users.
+
+### An hybrid approach
```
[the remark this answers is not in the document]
msbrogli: Ok.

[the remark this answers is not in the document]
raul-oliveira: I was my recommendation but I've updated the proposal to have a fee calculated by transaction size + priority tip (optional).

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+TODO
+
+# Drawbacks of the Hybrid Fee Model
+[drawbacks]: #drawbacks
+
+While the hybrid fee model balances incentives and network sustainability, it also introduces potential drawbacks that must be considered. Below are key concerns and challenges:
+
+### 1. Token Inflation and Supply Growth
+
+**Issue:** Since there is no burning mechanism tied to fee-based token transfers, the total supply of tokens may continuously increase without a natural deflationary counterbalance.
```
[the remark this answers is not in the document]
raul-oliveira: I removed this topic. It will depend from which destination we'll give to fees

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
msbrogli: I can't find the decision on what happens if a transaction has more than one fee-based token.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,232 @@
...
+
+**Impact:** Over time, token supply inflation can reduce the value of individual tokens, especially if demand does not grow proportionally.
+
+**Potential Mitigation:**
```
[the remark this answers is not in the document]
raul-oliveira: Makes sense. Removed this piece

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
msbrogli: I guess this part was removed.

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
msbrogli: I guess this part was removed.

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
msbrogli: I guess this part was removed.

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
msbrogli: Ok.

text/0045-transaction_fee_minting.md
msbrogli: Thanks!

text/0045-transaction_fee_minting.md
msbrogli: Ok.

text/0045-transaction_fee_minting.md
msbrogli: Please add this detail to the design.

text/0045-transaction_fee_minting.md
msbrogli: Please add this detail to the design.

text/0045-transaction_fee_minting.md
msbrogli: I guess you should. You can add a comment highlighting that the `tx_len = len(bytes(tx))` is variable and depends on the number of inputs and outputs. In the future, it will also depend on the number of headers.

text/0045-transaction_fee_minting.md
obiyankenobi: @hBulldozer replying your comment here so we can keep it in a thread.

> Thinking about it more now, I see that it seems complex

There's not much complexity in development, but on the product/business side. It may confuse some people, especially if it's not something that ends up being used regularly.

> A project could create tokenA that they control the supply and distribution of and can sell for a fixed price for example. And another tokenB that would be the transaction token for their service, so for users to use the service related to tokenB, they would need to have tokenA.

It's still not clear to me when this situation may happen. Do you have a concrete example that you've though of?

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Fee-Based Model
+
+Adding the fee-based model the platform won't require an upfront deposit. Instead, each transfer of the minted tokens incurs a transaction fee, which is deducted automatically. The exact fee rate will be defined at a later stage.
+
+By selecting the appropriate model, token creators can optimize their minting strategy based on their specific needs and usage scenarios. A key use case is the creation of memecoins.
+
+### How to approach
+
+An alternative to Bitcoin and Ethereum approchaes for blockchain fees is a hybrid model that combines fees based in the transaction size, with priority tip payments.
```
msbrogli, line 37: It's not clear to me how the priority tip would work in our case since we have a DAG of transactions and miners cannot freely choose which transactions will be confirmed by them.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,170 @@
...
+    outputs_size = 0
+    for output in tx.outputs:
+        outputs_size += 7 + len(output.script)
+        # 4 bytes (value) + 1 byte (token_data) + 2 bytes (script length prefix) + len(script)
+
+    # Total transaction size
+    tx_size = fixed_size + tokens_size + inputs_size + outputs_size
+
+    # Fee calculation
+    fee = tx_size * fee_rate_per_byte
```
msbrogli, line 83: What's the proposed value for the `fee_rate_per_byte`? It might be useful to get a histogram of tx size of transactions with custom tokens in mainnet so we can understand how much fee would be charged.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,170 @@
...
+
+Calculate size:
+
+| Field                | Calculation                 | Size     |
+|----------------------|-----------------------------|----------|
+| Fixed-size fields    |                             | **84** bytes |
+| Tokens               | 1 token × 32 bytes          | **32** bytes |
+| Inputs               | 2 × (35 + 108)              | **286** bytes |
+| Outputs              | 2 × (7 + 25)                | **64** bytes |
+| **Total Tx Size**    | 84 + 32 + 286 + 64          | **466** bytes |
```
msbrogli, line 105: This example is focused on calculating the tx size. But that's something we already have. I guess it would be more useful to have examples of the expected fee charged for small, medium and big transactions.

msbrogli (dismissed):

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,170 @@
...
+By selecting the appropriate model, token creators can optimize their minting strategy based on their specific needs and usage scenarios. A key use case is the creation of memecoins.
+
+### How to approach
+
+An alternative to Bitcoin and Ethereum approchaes for blockchain fees is a hybrid model that combines fees based in the transaction size, with priority tip payments.
+
+This model offers several advantages. First, it simplifies the user experience by providing a predictable fee without major fluctuations. Additionally, it enables a filtering mechanism to prevent excessively large or complex transactions, protecting the network from spam. Finally, its structure promotes sustainability, ensuring controlled blockchain growth.
+
+### Minning
+By adding fees to the transactions, we don't need to mine them anymore. So the proof of work (PoW) won't affect the fee calculation.
```
msbrogli, line 42: I feel we should keep the mining requirement for now. If I got it right, the fees will not be enough to prevent spam so we might need to keep this other mechanism in place.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,170 @@
...
+
+By selecting the appropriate model, token creators can optimize their minting strategy based on their specific needs and usage scenarios. A key use case is the creation of memecoins.
+
+### How to approach
+
+An alternative to Bitcoin and Ethereum approchaes for blockchain fees is a hybrid model that combines fees based in the transaction size, with priority tip payments.
+
+This model offers several advantages. First, it simplifies the user experience by providing a predictable fee without major fluctuations. Additionally, it enables a filtering mechanism to prevent excessively large or complex transactions, protecting the network from spam. Finally, its structure promotes sustainability, ensuring controlled blockchain growth.
+
+### Minning
```
msbrogli, line 41: Rename to Transaction Mining (so we don't confuse it with Block Mining).

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,170 @@
...
+By adding fees to the transactions, we don't need to mine them anymore. So the proof of work (PoW) won't affect the fee calculation.
+
+### Fee calculation
```
msbrogli, line 44: I guess it is still missing what's going to happen with the fee. Is it going to be burned? An example showing a concrete transaction would help a lot.

msbrogli:
Thanks @hBulldozer for joining us here.

I guess we might be mixing different things in the discussion.

Both Ethereum and Bitcoin have unpredictable fees because they use fees as a flow control mechanism. In their case, the calculate fees are the minimum fees and users can add more fees if they want their transactions prioritized. This prioritization process is exactly what makes fees unpredictable and so bad for their communities.

(I can't finish this message now, but I'll get back later today)

text/0045-transaction_fee_minting.md
hBulldozer: I don't have an example of a Protocol using this approach, but I can elaborate on a hypothetical scenario.
Let's say I want to operate stablecoins on the Hathor Network. However, I want to charge fees when users transact with my coins within the network. If I create a gas coin using the 1% HTR lock method (let's call it SGC—StableGasCoin), I can mint and sell it to users who will need it to pay transaction fees for all my stablecoins that are programmed to use SGC instead of HTR for gas.

The process would work like this:
1. Protocol mints SGC (StableGasCoin);
2. Protocol issues stablecoins with this rule - for example, mints sBRL with gas set to SGC instead of HTR;
3. Protocol sells SGC at arbitrary or fixed price (could be via a smart contract);
4. Users pay fees using the protocol's gas coin, initially minted using HTR.

This creates utility for the SGC coin and potential revenue for the protocol while also contributing to Hathor, as the 1% lock is burned.

msbrogli:
Returning to my argument, Bitcoin even allows transactions with zero fees. While there are no network restrictions against this, most miners discard zero-fee transactions from their mempool.

I don’t think flat fees are necessary to prevent fee unpredictability. Their downside is that they would reduce demand for HTR and require us to determine a single flat fee that works for both simple and complex transactions.

On another note, I agree that users should be able to pay fees using their deposit-based tokens in addition to HTR. This would significantly improve UX, as users wouldn’t need to hold HTR directly. I also support @trondbjoroy’s suggestion that wallets should prioritize HTR for fees by default. If no HTR is available, users should be able to choose which deposit-based token to use.

I don’t yet have a strong opinion on the best approach, but another alternative would be to charge fees based on the number of outputs with fee-based tokens. This would remove the incentive for bundling multiple transactions to save on fees.

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
msbrogli: I agree. So, let's assume that fees will be paid by burning tokens. More specifically, `fee = sum(outputs) - sum(inputs)`. Just clarify that another mechanism (e.g., miners collecting fees) can be put in place if needed.

text/0045-transaction_fee_minting.md
obiyankenobi: Got it, it's clearer now. It seems the use case would be to monetize a product/protocol that has usage.

I'm wondering if it makes sense to add this or just keep it simple and only accept HTR. Just to be clear, I think we should keep it simple for business reasons, not because development would be (slightly) harder.

trondbjoroy:
I agree with @msbrogli that a per-output based fee structure could make sense. It would align costs with transaction complexity while keeping it simple and predictable with no complex fee calculations, and it disincentivizes bundling. Since a solution like this would incur a fixed cost for each fee-based token output, and the total fee would scale linearly with the number of outputs, it might be more accurate to call it fixed fee rather than flat fee.

text/0045-transaction_fee_minting.md
msbrogli: > Almost all txs that a user sends have 2 outputs, as it needs to include a change output. So if we say it's 0.01 HTR per fee-based token output, it will cost 0.02. However, he may send a tx and the output match the input, so it will only cost 0.01 HTR. This discrepancy is not easy to explain to the average user.

It is very complicate to explain to EVM users how fees are calculated. But they all understand how much it cost. I guess it is the same here. One transaction would pay 0.01 HTR while the other would pay 0.02 HTR in fees. It seems simple enough for users.

> I guess only a few players would have the ability to do that, as exchanges and marketplaces. Most commonly, people will not do that.

Everything is hard to reach masses without tools. All we need is a popular wallet to implement it and then lots of people would use it unconsciously. For instance, mixers were theory before Tornado Cash made it simple for people to use.

obiyankenobi: > Everything is hard to reach masses without tools. All we need is a popular wallet to implement it and then lots of people would use it unconsciously.

I agree, but to reach this point there must be use on the network, and simplifying fees, which is a barrier to adoption, is a way to be able to get there.

> All we need is a popular wallet to implement it and then lots of people would use it unconsciously.

It will be a good problem when we reach the point when we have these kind of services in the network.

> It is very complicated to explain to EVM users how fees are calculated. But they all understand how much it costs. I guess it is the same here.

I don't think they understand how it works, they just pay what's displayed on the wallet. To quote @trondbjoroy "Bitcoin and Ethereum shouldn't be viewed as success stories when it comes to fees IMO". They do it because they want to send transactions, because there's demand for it. As a smaller network having to attract users and products, we need to look for enhancements to the current models.

text/0045-transaction_fee_minting.md
raul-oliveira: Added

text/0045-transaction_fee_minting.md
raul-oliveira: added

text/0045-transaction_fee_minting.md
raul-oliveira: I edited the explanation, pls check if it's make sense now.

text/0045-transaction_fee_minting.md
raul-oliveira: I'll grab it. For now I've added an example using the tx anamoty as reference. Pls, check if it's clear this way.

text/0045-transaction_fee_minting.md
raul-oliveira: Edited

text/0045-transaction_fee_minting.md
raul-oliveira: Removed, lets close this only with fee. I wasn't aware of this point. Tks

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,170 @@
...
+
+Adding the fee-based model the platform won't require an upfront deposit. Instead, each transfer of the minted tokens incurs a transaction fee, which is deducted automatically. The exact fee rate will be defined at a later stage.
+
+By selecting the appropriate model, token creators can optimize their minting strategy based on their specific needs and usage scenarios. A key use case is the creation of memecoins.
+
+### How to approach
+
+An alternative to Bitcoin and Ethereum approchaes for blockchain fees is a hybrid model that combines fees based in the transaction size, with priority tip payments.
+
+This model offers several advantages. First, it simplifies the user experience by providing a predictable fee without major fluctuations. Additionally, it enables a filtering mechanism to prevent excessively large or complex transactions, protecting the network from spam. Finally, its structure promotes sustainability, ensuring controlled blockchain growth.
```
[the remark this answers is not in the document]
raul-oliveira: Edited

text/0045-transaction_fee_minting.md
raul-oliveira: Edited

text/0045-transaction_fee_minting.md
raul-oliveira: I've changed the example, pls check if it's clear. I added a simple transaction with 1 input and two outputs. The idea is to become closer as possible to a flat fee in simple transaction.

text/0045-transaction_fee_minting.md
glevco: I think it's missing an explanation in the text on why it's useful for memecoins

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,198 @@
...
+          inputs_size += 35 + len(input.data)  
+          # 32 bytes (tx_id) + 1 byte (index) + 2 bytes (data length prefix) + len(data)
+    
+    # Outputs size calculation
+    outputs_size = 0
+    if len(tx.outputs) > 2:
+        for i in range(len(tx.outputs) - 2):
+            output = tx.outputs[i]
+            outputs_size += 7 + len(output.script)
+            # 4 bytes (value) + 1 byte (token_data) + 2 bytes (script length prefix) + len(script)
```
glevco, line 96: The output value size is variable, it can be 4 or 8 bytes depending on the value. Also, the final calculated fee value will affect at least one output value of the transaction, so to prevent this "recursion" I would exclude the output value from the fee calculation. This is analogous to how the input data is excluded from the sighash data of the tx (otherwise the signature would have to sign itself).

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,198 @@
...
+
+Calculate fee:
+
+| Field                | Calculation                 | Size          | Byte type | Total
+|----------------------|-----------------------------|---------------|-----------|-----|
+| Fixed-size fields     |                             | **84** bytes  | static    | 0.01 HTR
+| Tokens               | 1 token × 32 bytes          | **32** bytes  | dynamic   | 0 HTR  
+| Inputs               | 2 × 35                      | **286** bytes | dynamic   | 0.0038 HTR
+| Outputs              | 2 × (7 + 25)                | **64** bytes  | dynamic   | 0 HTR
+| **Total Fee**    |           |  |    | 0.0138 HTR
```
glevco, line 129: Since our network currently only supports up to 2 decimal places, the values used in this example are not practical. If we're going to use such specific low values, we would need some rounding mechanism, at least.

Another option would be suggesting an increase of decimal places in our network, but that's a big change that would have to be discussed independently. I guess all will depended on how the business decide to approach the fee sizes. Therefore, I guess you could use simpler values with only 2 decimal places in this example, and briefly mention this upgrade possibility if finer grained fees are desired.

text/0045-transaction_fee_minting.md
raul-oliveira: Done

text/0045-transaction_fee_minting.md
raul-oliveira: Done

text/0045-transaction_fee_minting.md
raul-oliveira: Changed the proposal, please check it again.

text/0045-transaction_fee_minting.md
```diff
@@ -0,0 +1,198 @@
...
+            # 4 bytes (value) + 1 byte (token_data) + 2 bytes (script length prefix) + len(script)
+
+    # Total transaction size
+    flat_fee = fixed_size * fee_rate_per_static_byte
+
+    ## Here we can also define how we want to calculate the dynamic fee. For example, imagine that we only want to charge fee for transactions that have more than two outputs, or ignore how the inputs size on the formula in order to make it predictable according to business needs.
+    dynamic_fee = (tokens_size + inputs_size + outputs_size) * fee_rate_per_dynamic_byte
+
+    # Fee calculation
+    fee = flat_fee + dynamic_fee + tx.priority_tip
```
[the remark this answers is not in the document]
raul-oliveira: Removed, doesn't make sense anymore.

text/0045-transaction_fee_minting.md
raul-oliveira: Removed. 
Changed the proposal, please check it again.

text/0045-transaction_fee_minting.md
glevco: @raul-oliveira as explained in person I was expecting some details on why this feature is useful for memecoins. This is just for documentation/reference purposes.

text/0045-transaction_fee_minting.md
glevco: Closing this as this sections was removed.

text/0045-transaction_fee_minting.md
glevco: @raul-oliveira as explained in person this section is still missing mentions to the other services that need to be updated.

text/0045-transaction_fee_minting.md
glevco: Closing this as this section was removed.

text/0045-transaction_fee_minting.md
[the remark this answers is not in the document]
glevco: Closing this as this section was removed.

text/0045-transaction_fee_minting.md
glevco: Closing this as this section was removed.

text/0045-transaction_fee_minting.md
raul-oliveira: Done. Check the clarify pls

glevco: Thanks!

text/0045-transaction_fee_minting.md
glevco: Thanks!

text/0045-transaction_fee_minting.md
raul-oliveira: Done

PR #103, opened 2026-02-25
feat: Integration test helper

tuliomir:
The Integration Test Helper is a dockerized HTTP service that provides race-condition-free wallet generation and funding for hathor-wallet-lib integration tests.

It pre-generates BIP39 wallets with derived addresses outside the test runtime (avoiding prohibitive address derivation costs under Jest) and splits the genesis block's single UTXO into a pool of independent, readily-available UTXOs so that parallel test executions never compete for funds.

See [rendered](https://github.com/HathorNetwork/rfcs/blob/feat/integration-test-helper/projects/hathor-wallet-lib/0000-integration-test-helper.md).

### DIFF
--- a/projects/hathor-wallet-lib/0000-integration-test-helper.md
+++ b/projects/hathor-wallet-lib/0000-integration-test-helper.md
@@ -1,0 +1,436 @@
+- Feature Name: integration-test-helper
+- Start Date: 2026-02-24
+- Author: Tulio <tulio@hathor.network>)
+
+# Summary
+
+The [Integration Test Helper](https://github.com/tuliomir/hathor-integration-test-helper)
+is a dockerized HTTP service that provides race-condition-free wallet
+generation and funding for
+[hathor-wallet-lib](https://github.com/HathorNetwork/hathor-wallet-lib)
+integration tests. It pre-generates BIP39 wallets with derived addresses
+outside the test runtime (avoiding
+[prohibitive address derivation costs under Jest](https://github.com/HathorNetwork/hathor-wallet-headless/issues/173))
+and splits the genesis block's single UTXO into a pool of independent,
+readily-available UTXOs so that parallel test executions never compete
+for funds.
+
+# Motivation
+
+The [hathor-wallet-lib](https://github.com/HathorNetwork/hathor-wallet-lib)
+integration test suite currently has two structural blockers that prevent
+test parallelization:
+
+## 1. Wallet generation is prohibitively slow under Jest
+
+Deriving addresses from BIP39 seeds under Jest's execution wrapper takes
+7 seconds for simple wallets and up to 60 seconds for multisig wallets,
+compared to under 500 ms when running outside Jest
+([hathor-wallet-headless#173](https://github.com/HathorNetwork/hathor-wallet-headless/issues/173)).
+The current workaround is a
+[static JSON file with 100 pre-calculated wallets](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/__tests__/integration/helpers/wallet-precalculation.helper.ts)
+that are consumed serially. This works for sequential execution but
+breaks under parallelization: concurrent tests drawing from the same
+file race to consume wallets, leading to duplicates and test failures.
+
+## 2. All funding derives from a single genesis UTXO
+
+Every integration test that needs funds must ultimately spend from the
+same genesis block UTXO. When tests run in parallel, multiple funding
+transactions attempt to spend the same UTXO simultaneously, causing
+race-condition failures. There is no built-in mechanism in
+[hathor-wallet-lib](https://github.com/HathorNetwork/hathor-wallet-lib)
+to coordinate UTXO selection across independent test processes.
+
+## Expected outcome
+
+With this service running alongside the test infrastructure, integration
+tests can:
+
+- Instantly obtain unique, ready-to-use wallets with pre-derived
+  addresses (no Jest-side derivation).
+- Request funding without coordinating with other tests — each request
+  gets its own dedicated UTXO, eliminating race conditions.
+- Run fully in parallel, reducing total suite execution time
+  proportionally to the number of parallel workers.
+
+# Guide-level explanation
+
+## Overview
+
+The Integration Test Helper is a lightweight HTTP service that runs as a
+Docker container alongside the existing test infrastructure
+([hathor-core](https://github.com/HathorNetwork/hathor-core) fullnode +
+[tx-mining-service](https://github.com/HathorNetwork/tx-mining-service)).
+It exposes four endpoints:
+
+| Method | Path             | Purpose                                    |
+|--------|------------------|--------------------------------------------|
+| GET    | `/simpleWallet`  | Get a new simple wallet (seed + addresses) |
+| GET    | `/multisigWallet`| Get N-of-M multisig wallets                |
+| POST   | `/fund`          | Fund a test wallet address with HTR        |
+| GET    | `/status`        | Pool health and readiness                  |
+
+## Wallet generation
+
+When a test needs a wallet, it makes a `GET /simpleWallet` request. The
+service returns a 24-word BIP39 seed and 22 pre-derived addresses. Since
+address derivation happens inside the service (not under Jest), it
+completes in milliseconds. Wallets are pre-generated and cached, so
+responses are near-instant.
+
+For multisig tests, `GET /multisigWallet?participants=N&numSignatures=M`
+generates N participant wallets that share the same P2SH addresses.
+
+## Funding
+
+When a test needs HTR, it makes a `POST /fund` request with the target
+address and optional amount. The service atomically reserves a UTXO from
+its internal pool, builds and broadcasts a transaction, and returns the
+transaction ID. No two tests ever attempt to receive funds from the same UTXO.
+
+## Typical test workflow
+
+```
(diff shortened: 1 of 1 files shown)


projects/hathor-wallet-lib/0000-integration-test-helper.md
```diff
@@ -0,0 +1,439 @@
...
+2. Otherwise: queue the request with a timeout
+   (`FUND_TIMEOUT_MS`, default 30s)
+3. Request resolves when a large UTXO becomes available (e.g., after
+   a refill), or times out with HTTP 409
+
+### Auto-refill
+
+When `testUtxos.length` drops below `REFILL_THRESHOLD` (default 10)
+after a funding operation, the service triggers a background split of
+the current `largeUtxo`. The split is delayed by 1.5 seconds to avoid
```
r4mmer, line 312: This wait X seconds for the tx to be accepted usually introduces race conditions, a solution we have on the wallet-lib itself is waiting for both wallets to receive the transaction, this way we always wait until it is ready to use again.

projects/hathor-wallet-lib/0000-integration-test-helper.md
```diff
@@ -0,0 +1,439 @@
+- Feature Name: integration-test-helper
```
r4mmer, line 1: __Unrelated to line__

I got a couple os suggestions for the service:

1)
Define a consistent error response schema, e.g.: `{ "error": "POOL_EXHAUSTED", "message": "No UTXOs available", "retryable": true }`
This creates possible recoverable actions like retry and makes the reason of failure easier to spot.

2)
Increate log observability with test names, meaning that on all requests we send the test name and the service logger will include the test name on the logs.
This way we can search for the test name and get the logs from their requests, this would make debugging any issues easier in the future

tuliomir: Good points, thanks!

1. Added an error schema on https://github.com/HathorNetwork/rfcs/pull/103/commits/3357814fb1f4efd82ee276dec0fdce23b942a54b

2. Added the test name feature on https://github.com/HathorNetwork/rfcs/pull/103/commits/3006b32cb9cd7002d4f8c5f0cc14558647e6e5a4

projects/hathor-wallet-lib/0000-integration-test-helper.md
tuliomir: The idea of having a buffer of transactions is exactly to prevent race conditions: the tests have another 1.5 seconds to request more transactions while the buffer is re-filled.

It is not necessary to wait for the tx to be received by the wallet, because the utxo is guaranteed to be isolated for that test alone.

If this becomes not enough we can increase the threshold to 20, for example. Another option, if the DevMiner ( #102 ) becomes efficient enough, is to reduce the delay from 1500ms to less than that.

r4mmer (dismissed):

projects/hathor-wallet-lib/0000-integration-test-helper.md
```diff
@@ -0,0 +1,509 @@
...
+└───────────────────────────────────────┘
+    │
+    ▼ needsRefill() triggers when testUtxos < threshold
+┌───────────────────────────────────────┐
+│  splitUtxo(largeUtxo) → 100 more     │
+│  testUtxos + new largeUtxo           │
+└───────────────────────────────────────┘
+```
+
+### Race-condition freedom
```
pedroferreira1, line 215: **Re: race-condition freedom claim — there's a hidden wallet-lib coupling to call out**

The reference-level explanation states:
> The pool achieves race-condition freedom through JavaScript's single-threaded event loop. `reserveUtxo()` performs a synchronous array `shift()` — the dequeue completes atomically before any other request handler can execute.

This is correct for the dequeue itself, but `fundAddress` doesn't end at the dequeue — it spans two async boundaries (build/sign and `runFromMining`), and during those yields any other handler (including `rescanUtxoPool`) can run. Whether the pool stays consistent across those yields depends on **wallet-lib internals that the RFC doesn't acknowledge**, and the assumptions don't hold under load.

**What I traced in `hathor-wallet-lib`:**

Two mechanisms hide a UTXO from `getAvailableUtxos()`:

1. `utxosSelectedAsInput` — an in-memory `Map<"txId:index", boolean>` on `Storage` ([`src/storage/storage.ts#L77`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/storage/storage.ts#L77)). The filter that excludes "selected" UTXOs is in `Storage.selectUtxos` ([`src/storage/storage.ts#L521-L549`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/storage/storage.ts#L521-L549)).
2. `spent_by` on the source output — set asynchronously when the wallet processes the new tx via `enqueueOnNewTx`.

The transient lock (1) is set inside `mineTx()` via `updateOutputSelected(true)` ([`src/new/sendTransaction.ts#L676-L681`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/new/sendTransaction.ts#L676-L681)) — that is, **at the start of mining, not at signing**. Importantly, it carries a TTL of `SELECT_OUTPUTS_TIMEOUT = 1000 * 60` ms ([`src/constants.ts#L249`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/constants.ts#L249)) and `utxoSelectAsInput` schedules a `setTimeout` that auto-releases the lock after 60s ([`src/storage/storage.ts#L763-L788`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/storage/storage.ts#L763-L788)), regardless of whether the tx has been observed by the wallet's WS pipeline.

**Two windows where the helper's pool view can disagree with `getAvailableUtxos()`:**

*Window A — between `reserveUtxo` (sync) and `mineTx` (first await inside `runFromMining`):*
```
const utxo = reserveUtxo(amount);              // OUT of helper bucket
const tx   = await buildAndSignWithTemplate(); // ← yield, tx signed but NOT yet marked
await sendTx.runFromMining();                  // ← updateOutputSelected(true) runs here
```
During this window the UTXO is out of the helper bucket but still visible in `getAvailableUtxos()`. If a concurrent request triggers `rescanUtxoPool()` here (e.g., on a stale-UTXO error), rescan repopulates `testUtxos` from `getAvailableUtxos()` and reintroduces the in-flight UTXO. The next reservation hands the same UTXO to a second request.

*Window B — successful broadcast → WS observation, bounded by 60 s:*
After `runFromMining` resolves, the broadcast's success is the JSON response from `/push_tx`; `enqueueOnNewTx` is fire-and-forget at [`src/new/sendTransaction.ts#L767`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/new/sendTransaction.ts#L767), and `spent_by` only flips once the wallet has processed the new tx (typically via WS event). Until then, the *only* thing keeping the UTXO out of `getAvailableUtxos()` is the transient lock — and that lock expires in 60 s. On a CI runner with parallel tests, bursty PoW, and WS lag, breaching 60 s isn't theoretical; once breached, a rescan re-adds the just-spent UTXO to the pool, the next request reserves it, and `runFromMining` fails with "already been spent" — back at the brittle string match.

**Why this should change the RFC, not just the implementation:**

The current spec leans on `utxosSelectedAsInput` and the 60 s `SELECT_OUTPUTS_TIMEOUT` constant without naming them. Both are private wallet-lib internals — neither is part of any public contract, and either could change in a minor version bump and silently break the helper's correctness. Two requests:

1. **Add an explicit invariants subsection** to the reference-level explanation. Suggested wording:
   > A UTXO returned by `reserveUtxo` is added to a helper-owned `reservedSet` in the same synchronous tick as the dequeue. `rescanUtxoPool` MUST exclude any UTXO in `reservedSet` when repopulating buckets. Cleanup happens in a `finally` only after the consuming tx is observed as confirmed (e.g., the parent output's `spent_by` is set), not merely after `runFromMining` returns. This makes correctness independent of `utxosSelectedAsInput`, `SELECT_OUTPUTS_TIMEOUT`, the mine-vs-sign timing, and WS propagation latency — all wallet-lib internals.

2. **Add a "Dependencies on wallet-lib internals" note** listing the names above and stating that the helper's design no longer relies on them once `reservedSet` is in place — only on the public `selectUtxos`/`runFromMining`/`enqueueOnNewTx` API surface.

Optional but cheap: an option to disable wallet-lib's TTL auto-release for the helper's session (or to configure a much larger value), as a belt-and-braces against bugs in the helper's own bookkeeping.

pedroferreira1 (dismissed):

projects/hathor-wallet-lib/0000-integration-test-helper.md
tuliomir: Thanks for the careful trace — you're right that the `reserveUtxo` atomicity claim glossed over the async boundaries inside `fundAddress`, and that the helper was implicitly relying on `utxosSelectedAsInput` and `SELECT_OUTPUTS_TIMEOUT`.

Adopted both suggestions:

- RFC: replaced the "Race-condition freedom" section with explicit invariants (`reservedSet`, rescan exclusion, observed-then-released cleanup, bounded fallback) and added a "Dependencies on wallet-lib internals" subsection. See b487ea3.
- Implementation: tuliomir/hathor-integration-test-helper#2.

pedroferreira1 (approved):

r4mmer (approved):

PR #104, opened 2026-03-03
design: Shielded outputs

msbrogli:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/feat/shielded-outputs/text/0000-shielded-outputs.md)


text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,884 @@
...
+3. **Nonce derivation**: `nonce = SHA256("Hathor_CT_nonce_v1" || s)`.
+4. **Range proof rewind**: Using the nonce as the rewind key, the wallet calls `rewind_range_proof()` which returns the committed `(value, blinding_factor, message)`.
+5. **For `FullShieldedOutput`**: The `message` field contains `token_uid(32B) || asset_blinding_factor(32B)`. The wallet reconstructs the expected asset commitment from these values and cross-checks against the on-chain `asset_commitment` to prevent a malicious sender from claiming a worthless token is HTR.
+
+If rewind fails (wrong nonce), the output is not addressed to this wallet — skip it silently.
+
+### Sending Shielded Payments
+
+1. **Choose privacy tier** per output (`AmountShieldedOutput` or `FullShieldedOutput`).
+2. **Generate ephemeral keypair** `(e, E = e*G)` for each output.
```
obiyankenobi, line 262: How do you generate it? Do you take some info from the output to make it unique?

msbrogli: No, you generate a random number `e` and multiply by `G`.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+This transparency creates real problems:
+
+- **Payroll and compensation.** An employer paying salaries on-chain reveals every employee's compensation to anyone who looks.
+- **Business-to-business payments.** Suppliers, vendors, and partners can reverse-engineer pricing, margins, and deal terms from on-chain flows.
+- **Trading and DeFi.** Front-runners and MEV extractors exploit visible amounts to sandwich trades or copy strategies.
+- **Personal finance.** Any recipient of a payment can trace the sender's full balance and transaction history.
+- **Multi-token privacy.** Hathor supports custom tokens. When token types are visible, observers can track the flow of specific assets (e.g., loyalty points, governance tokens, stablecoins), revealing business relationships and portfolio composition.
+- **Confidential issuance.** A token issuer who routinely transacts in shielded form should be able to mint or melt without dropping to fully-transparent mode, while still letting anyone audit total supply.
```
LFRezende, line 33: Not wrong, but I think you're hinting far too early the concept of shielded minting/melting, and not necessarily explaining why we need a shielded issuancer to cover for transparent minting/melting.

Maybe something on the lines of "while transparency for minting/melting operations is fundamental to probe the network's balance of token supplies, the transparency of these operations reveals the transaction that issued the supply change, exposing the identity of the issuer."

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+Hathor offers **two privacy tiers**, selectable per output:
+
+| Tier | Hides Amount | Hides Token | Proof Overhead | Use Case |
+|------|:---:|:---:|---|---|
+| `AmountShieldedOutput` | Yes | No | Range proof (~3213 B) | Hide salary amounts while token type is public |
+| `FullShieldedOutput` | Yes | Yes | Range proof + surjection proof (~3213 + variable B) | Full privacy for multi-token transactions |
+
+Both tiers use **Pedersen commitments** for amounts and **Borromean range proofs** to guarantee amounts are non-negative (specifically in `[1, 2^40)`). `FullShieldedOutput` additionally uses **blinded asset tags** and **asset surjection proofs** to hide and validate the token type.
+
```
LFRezende, line 67: It is increasingly clear we will not go to mainnet with Borromean, both for cost optimizations and to parallelize verifications and proofs. So we need to change this section to include Bulletproofs.

```md

| Tier | Hides Amount | Hides Token | Proof Overhead | Use Case |
|------|:---:|:---:|---|---|
| `AmountShieldedOutput` | Yes | No | Range proof (~ $675 + 64\cdot log_2 (n)$ bytes) | Hide salary amounts while token type is public |
| `FullShieldedOutput` | Yes | Yes | Range proof (same as above) + surjection proof (variable bytes) | Full privacy for multi-token transactions |

```

Additionally, the range of the proof is not 
```md 
$[1, 2^40]$ 
``` 
But: 
```md 
$[1, 2^(64 + \ceil{\frac{D}{log_10 2}})]$ 
``` 

Where `D` is the decimals we intend to add in the future as precision. 

For simplicity, we can say for now  ` [1, 2^k] `, where `k` is the number of range proof bits we intend to use, where typically `k == 64`.

Final paragraph rewritten:

```markdown

| Tier | Hides Amount | Hides Token | Proof Overhead | Use Case |
|------|:---:|:---:|---|---|
| `AmountShieldedOutput` | Yes | No | Range proof (~ $675 + 64\cdot log_2 (n)$ bytes) | Hide salary amounts while token type is public |
| `FullShieldedOutput` | Yes | Yes | Range proof (same as above) + surjection proof (variable bytes) | Full privacy for multi-token transactions |

Both tiers use **Pedersen commitments** for amounts and **Bulletproof range proofs** to guarantee amounts are non-negative (specifically in `[1, 2^k)`), where k is the number of range proof bits we intend to use, typically `k == 64`. `FullShieldedOutput` additionally uses **blinded asset tags** and **asset surjection proofs** to hide and validate the token type.
```

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+|  ECDH-derived one-time addresses                          |
+|  "Who received it?" -> Hidden                             |
++-----------------------------------------------------------+
+```
+
+Each phase is independently useful and composes with the others. With all three active, an observer learns almost nothing about a transaction beyond the number of inputs and outputs.
+
+## 2. How It Works
+
+### Shielding: Moving Funds into Privacy
```
LFRezende, line 94: Opinion: The shielding section basically assumes profficiency with pedersen commitments, blinding factors and such. 

Adding a "preliminaries" subsection here to introduce the basic concepts and hooks we'll work with, such as pedersen commitments, hash-to-point and such,  could do massive work to easen the blow to newcomers.

No need to go too technical, just introducing and explaining what it does is enough.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+```
+FULL-UNSHIELD TRANSACTION:
+  Input:   [commitment] (shielded, 90 HTR hidden)
+  Output0: 85 HTR to Bob (transparent)
+  Fee:     5 HTR (transparent)
+
+  Header: UnshieldBalanceHeader { excess_blinding_factor (32B scalar) }
+
+Observer sees: 85 HTR left the shielded pool
+Observer does NOT know: the shielded input amount (other than that it equals 85 + 5)
```
LFRezende, line 133: Scratch that. He does know the amount if he can sum. The whole point is unshielding. If we wished him not to know, we'd have two shielded inputs.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+```
+
+> **Privacy footgun.** The scalar form of `UnshieldBalanceHeader` exposes
+> `sum(r_in)` of the unshielded inputs as a raw scalar. A point-form replacement
+> (`E_tx = excess·G` plus a Schnorr binding signature) closes this; it is **not
+> implemented** and is described in [Future possibilities](#future-possibilities).
+
```
LFRezende, line 140: I think we already have this implemented? Or if not, I think we create pedersen commitments with zero value and just the excess blinding factor.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+The declared amounts are bound into the homomorphic balance equation as public scalar terms (Rule M4), so the verifier enforces supply correctness without learning where the new value lands or which UTXO was melted. Because the amounts are public, total per-token supply remains computable by anyone. What becomes private is the *recipient set*, the *per-recipient amounts*, and *which shielded UTXO is melted*.
+
+A given token may appear in **either** `MintHeader` **or** `MeltHeader` in a single transaction, never both (Rule M3). Worked examples appear in [§2.7](#27-mintmelt-examples).
+
+### Fee Model
+
+Shielded outputs impose additional verification costs (range-proof verification, surjection-proof verification, larger storage). A per-output fee compensates the network:
+
+- `FEE_PER_AMOUNT_SHIELDED_OUTPUT`: charged per `AmountShieldedOutput` (default: `1` HTR base unit)
+- `FEE_PER_FULL_SHIELDED_OUTPUT`: charged per `FullShieldedOutput` (default: `2` HTR base units)
```
LFRezende, line 172: Is it `1 HTR`/`2 HTR`? I thought it was a higher base fee per output (wasn't it 10/20?)

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+The spend pubkey is represented as a 20-byte hash (like a standard P2PKH address). Shorter addresses, and fully sufficient for shielded outputs (ECDH uses the scan key; output locking uses the spend key hash). However, the full spend pubkey is not recoverable from the address alone, which prevents future Silent Payments integration (one-time address derivation requires EC point addition on `B_spend`).
+
+#### Option B: Full Keys (66 bytes of data)
+
+```
+scan_pubkey(33) || spend_pubkey(33)
+```
+
+Both keys are present in full. Longer addresses, but the sender has everything needed without any additional lookups. Compatible with BIP352-style Silent Payments, where the spend pubkey is required for one-time address derivation.
```
LFRezende, line 221: I had an idea - haven't given too much thought, but I leave it here recorded.

In order to reduce size, we could make the scan pubkey deductable from the spend pubkey. 

A third-party could fetch the STx's data (say, message `m`) and the available Spend Pubkey (`S`) to obtain the View PubKey (`V`). He then could do:

`t = H(m || S || ...) # taproot encoding message and the spend pubkey (and maybe some other stuff)`

`V = S + t*G # View key for that particular transaction`

Since no one knows the discrete log of S, no one is able to either recover the private spend key (`s = S * G^(-1)`), nor the private view key (`v = V*G^(-1)`), yet we can generate on the fly the view key, saving space and still allowing third party services to generate these keys programatically to aggregate them.

Additionally, the receiver B can easily do ECDH with the incoming transactions to check what sender A sent them.

```python

# sender A: 
t = H(m || S) # creates a taproot with the message of STx and the spend key

v = s + t # A knows its spend private key and the taproot is easily produceable

# v*B == s*B + t*B == S*b + t*B (ECDH) (*)

secret = H(v*B)  # This will be the secret prepped by A to B.

#####

# Now B recovers the secret

```python

# He fetches the message and the spend key and computes t
t = H(m || S)

# Then builds the public view key
V = S + t*G

# Finally rebuilds the secret
# V*b == S*b + t*b*G == S*b + t*B (ECDH) (Same as (*)).

`secret = H(V*b)`
```

Now B can rewind the proofs. Only he and the sender. 

With just 33 bytes.

If we wish to centralize further towards only one view key, regardless of transaction (to facilitate delegated scanning), we can do this:

`t = H(S) # Take away variants. Just the spend pub key. `
`V = S + t*G # View key fixed.`

So long as one does not need the viewing PRIVATE key (`v = s + t` makes it easy to get the spending private key), we should be fine.

We could even make something as simple as `v == H(S)*secret` and `V == secret*H(S)*G`, in this case the `secret = something else`, with the spend key.. View key easily derivable from S, does not expose spending private key, and easens the blow for delegated scanning services. 

Again, just thought of it, may be crude or flawed.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+### Receiving Shielded Payments
+
+Every shielded output contains an `ephemeral_pubkey` field (33 bytes; all-zero means "not present"). The recipient recovers the hidden amount through **ECDH + range proof rewind**:
+
+1. **Address match**: The wallet checks if the output script contains `hash(spend_pubkey)` matching one of its keys.
+2. **ECDH shared secret**: `s = ECDH(scan_privkey, ephemeral_pubkey)` (secp256k1 ECDH; SHA256 of the shared point).
+3. **Nonce derivation**: `nonce = SHA256("Hathor_CT_nonce_v1" || s)`.
+4. **Range proof rewind**: Using the nonce as the rewind key, the wallet calls `rewind_range_proof()` which returns the committed `(value, blinding_factor, message)`.
+5. **For `FullShieldedOutput`**: The `message` field contains `token_uid(32B) || asset_blinding_factor(32B)`. The wallet reconstructs the expected asset commitment from these values and cross-checks against the on-chain `asset_commitment` to prevent a malicious sender from claiming a worthless token is HTR.
```
LFRezende, line 255: The ephemeral pubkey is still in the pipeline? Or are we using the remainder (if any) of the blinding factors in the pedersen commitments as it?

The ephemeral pubkeys were phase C if I'm correct.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+### Sending Shielded Payments
+
+1. **Choose privacy tier** per output (`AmountShieldedOutput` or `FullShieldedOutput`).
+2. **Generate ephemeral keypair** `(e, E = e*G)` for each output.
+3. **Compute ECDH shared secret** with the recipient's scan public key (`B_scan`).
+4. **Derive nonce** for range proof construction.
+5. **Create Pedersen commitment**: `C = amount * H_token + blinding * G` (with `H_token` the blinded asset generator for `FullShieldedOutput`).
+6. **Create range proof** using the deterministic nonce (enables recipient rewind).
+7. **For `FullShieldedOutput`**: Create the blinded asset commitment and surjection proof, and embed `token_uid || asset_blinding_factor` in the range proof message.
```
LFRezende, line 267: I'm still ansy on it. The channel seems to be very limited in size (20 B) and the blinding factor alone is 32 B. Need to think on how to encode it, or if the 20 B is due to the lib implementation.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+When all outputs are controlled by one party (the common non-swap case), a single aggregated proof could reduce total proof size logarithmically, if an MPC-free aggregation path becomes available.
+
+## View key delegation for compliance
+
+A scan-key-derived view key would let a designated party (auditor, regulator, exchange) decrypt shielded outputs addressed to them without spending authority — analogous to Monero view keys.
+
+## Asset-balance binding signature
+
+The (future) binding signature covers value balance; a second one over asset balance would enable chain-wide *asset* audit for `FullShieldedOutput` transactions, detecting cross-asset forgery the same way inflation is detected today.
+
```
LFRezende, line 1117: Mimblewimble does this with pedersen commitments.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+The construction is exactly Sapling's binding signature (also used by Mimblewimble and Liquid CT): deterministic Schnorr over secp256k1 with a `"HathorBindingSig/v1"` domain separator, batch-verifiable across a block. Neither variant is in the current code; the network-wide supply audit ([§4.7](#47-network-wide-supply-audit)) already works without it.
+
+## Bulletproof range proofs
+
+Migrating from Borromean (~3213 B) to Bulletproof (~675 B) range proofs would cut shielded-output size ~5× and enable true batch verification, gated on `secp256k1-zkp` exposing a compatible Bulletproof + rewind API.
+
```
LFRezende, line 1101: Not a future possibility, but our current target.

We can perhaps put bulletproof+

Or even bulletproof++, using the argument that it is not being used currently due to being fairly new and with no production-grade implementations of it.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+1. **Shielded address format.** Option A (compact) vs Option B (full keys) vs Option C (spend-only). The tradeoff is address length vs. Silent Payments compatibility and light-client scanning. Should be resolved before wallet implementations begin. (Code currently assumes the full public key — Option B.)
+
+2. **Final fee amounts.** The placeholders (`FEE_PER_AMOUNT_SHIELDED_OUTPUT = 1`, `FEE_PER_FULL_SHIELDED_OUTPUT = 2`) are not final and should be set from measured verification-cost differentials and network economics.
+
+3. **Batch verification.** The current `batch_verify_range_proofs` is a sequential loop. A true batched multi-exponentiation API would reduce CPU for multi-output transactions; timeline depends on library support.
```
LFRezende, line 1075: Bulletproofs may allow us to scratch this out.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+## Monero (RingCT)
+
+Monero uses Pedersen commitments and range proofs for amount hiding plus ring signatures for sender privacy; all transactions are mandatory CT. Differences: Ed25519 (not secp256k1), mandatory (not opt-in), combined amount+sender privacy, no custom tokens. Lessons: range-proof CT is production-proven at scale; ECDH recovery with deterministic nonces is the standard approach.
+
+## Zcash (Sapling / Orchard)
+
+Zcash uses ZK-SNARKs for full transaction privacy in a single proof, on BLS12-381 / Pallas-Vesta. Sapling's `valueBalance` field and binding signature are the closest analogs to a per-asset public delta on an otherwise-shielded transaction (the inspiration for the deferred point-form excess). Differences: different curves, more expensive proving, separate shielded pool. Lesson: opt-in privacy reduces the anonymity set, so shielded usage should be encouraged.
```
LFRezende, line 1062: And also incapable of being audited or noting an inflation attack.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+## Why separate range proofs (not aggregated)?
+
+Aggregated proofs would require the prover to know all values and blinding factors for every output, which is incompatible with atomic swaps and multi-party transactions where each party independently constructs their outputs. Separate proofs also avoid permanently linking all outputs as same-party and allow independent UTXO pruning.
+
+## Why Borromean range proofs (not Bulletproofs)?
+
+The implementation uses `secp256k1-zkp`'s `RangeProof` (a Borromean-style proof). It is the proof type exposed by the library version in use, battle-tested in production (Liquid/Elements), and supports the ECDH-rewind recovery mechanism directly. The cost is size: ~3213 bytes vs a Bulletproof's ~675 bytes. Bit width is pinned to 40 so all proofs are constant-size, removing a value-size side-channel. Migrating to Bulletproofs is a future optimization gated on library support.
+
```
LFRezende, line 1020: We'll scratch that.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+3. **Wallet complexity.** Wallets must implement ECDH key exchange, range-proof rewind, blinding-factor management and balancing, surjection-proof generation, and fee computation — a substantial increase over transparent-only transactions.
+
+4. **Explorer capability reduction.** Explorers lose amount/balance display for shielded outputs (though per-token supply remains auditable from the mint/melt headers).
+
+5. **Recipient public key requirement.** The sender must know the recipient's full compressed public key (not just the address hash) to establish the ECDH shared secret. This requires a prior on-chain spend, a payment protocol, or a new address format.
+
+6. **Full-unshield scalar leak.** `UnshieldBalanceHeader` exposes `Σ r_in` as a raw scalar. A point-form replacement with a binding signature would close this but is not implemented (see Future possibilities).
+
+7. **40-bit value ceiling.** Range proofs prove `[1, 2^40)`. This is above the maximum HTR supply but is a smaller domain than the full `u64` amount space; any token whose supply could exceed `2^40 − 1` base units cannot be fully represented in a shielded output.
```
LFRezende, line 1000: Not 40 bit.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+A `NanoHeader` may coexist with mint/melt headers, subject to Rule M6 (no same-token supply change via both channels). Deeper interaction — contracts consuming or producing shielded outputs — is out of scope for this RFC.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+1. **Transaction size increase.** A shielded output is ~3.3 KB (dominated by the ~3213-byte Borromean range proof) vs ~40 bytes for a transparent output — roughly two orders of magnitude larger. This increases storage, bandwidth, and propagation time. The range proof is the single largest contributor; a Bulletproof-based proof (~675 B) would shrink this substantially but is not available in the current `secp256k1-zkp` build.
+
+2. **Verification cost.** Each range proof and each surjection proof is verified independently. Batch verification is not yet truly batched (the helper loops sequentially), so multi-output transactions pay close to linear cost.
+
```
LFRezende, line 991: Bulletproofs do batch and are per-transaction, so no helper-loops. We can scratch this paragraph.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+### Nano Contracts
+
+A `NanoHeader` may coexist with mint/melt headers, subject to Rule M6 (no same-token supply change via both channels). Deeper interaction — contracts consuming or producing shielded outputs — is out of scope for this RFC.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+1. **Transaction size increase.** A shielded output is ~3.3 KB (dominated by the ~3213-byte Borromean range proof) vs ~40 bytes for a transparent output — roughly two orders of magnitude larger. This increases storage, bandwidth, and propagation time. The range proof is the single largest contributor; a Bulletproof-based proof (~675 B) would shrink this substantially but is not available in the current `secp256k1-zkp` build.
+
```
LFRezende, line 989: Bulletproofs are now the target.

I think we can keep this paragraph but just mention the size will significantly increase to kB instead of mere bytes.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+7. Store `E` (33 bytes, compressed) in the output's `ephemeral_pubkey` field.
+
+### Recipient Flow
+
+1. Match the output script against a wallet key.
+2. Extract `E`; compute `s = ECDH(scan_privkey, E)` (equal to the sender's `s`).
+3. Derive `nonce = SHA256("Hathor_CT_nonce_v1" || s)`.
+4. Call `rewind_range_proof(proof, commitment, nonce, generator)` with `generator = derive_asset_tag(token_uid)` for `AmountShieldedOutput` or `output.asset_commitment` for `FullShieldedOutput`.
+5. Receive `(value, blinding_factor, message)`.
+6. For `FullShieldedOutput`: extract `token_uid` and `asset_blinding_factor` from `message`, recompute the expected asset commitment, and cross-check it against the on-chain `asset_commitment` (rejects a fraudulent token UID).
```
LFRezende, line 944: I think we can rewind the asset blinding factor without needing to unpack it from the message. The token uid I still have my questions.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+## 4.6 Fee Mechanism
+
+### Settings
+
+| Setting | Default | Location | Description |
+|---------|---------|----------|-------------|
+| `FEE_PER_AMOUNT_SHIELDED_OUTPUT` | 1 | `hathor/conf/settings.py` | HTR base units per `AmountShieldedOutput` |
+| `FEE_PER_FULL_SHIELDED_OUTPUT` | 2 | `hathor/conf/settings.py` | HTR base units per `FullShieldedOutput` |
+| `FEE_PER_OUTPUT` | 1 | `hathorlib` settings | HTR per chargeable transparent output and per `FEE`-token mint/melt entry |
+
```
LFRezende, line 897: Are we really going with 1 HTR and 2 HTR? Is it set in stone?

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+`verify_commitments_valid` — every commitment, asset commitment, and (present) ephemeral pubkey is a valid 33-byte secp256k1 point, and the output count is within `MAX_SHIELDED_OUTPUTS`. Raises `InvalidShieldedOutputError`. Standard transaction structure rules (≥1 input, ≥1 output) continue to apply.
+
+### Rule 2: Fee Is Always Transparent
+
+`verify_shielded_fee` — a shielded transaction MUST carry a `FeeHeader`. Fees enter the balance equation as trivial commitments. Raises `InvalidShieldedOutputError`.
+
+### Rule 3: Range Proofs
+
+`verify_range_proofs` — every shielded output MUST include a valid range proof for `[1, 2^40)` against its commitment and generator (unblinded for `AmountShieldedOutput`, blinded for `FullShieldedOutput`). Raises `InvalidRangeProofError`.
+
```
LFRezende, line 768: Not 40, again.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+|-------|------|------|-------------|
+| `commitment` | bytes | 33 B | Pedersen commitment (uses the blinded asset generator) |
+| `range_proof` | bytes | ~3213 B (max 3328) | Borromean range proof |
+| `script` | bytes | variable (max 1024) | Locking script |
+| `asset_commitment` | bytes | 33 B | Blinded asset tag `A = H_token + r_asset*G` |
+| `surjection_proof` | bytes | variable (max 4096) | Asset surjection proof |
+| `ephemeral_pubkey` | bytes \| None | 33 B | Compressed secp256k1 point for ECDH recovery (all-zero = not present) |
+
+**Typical total size**: ~3.3 KB per shielded output, dominated by the range proof.
+
```
LFRezende, line 641: Again, if 40 bits and Borromean. But if bulletproof, it is by transaction, and much less.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+Hides both amount and token type.
+
+| Field | Type | Size | Description |
+|-------|------|------|-------------|
+| `commitment` | bytes | 33 B | Pedersen commitment (uses the blinded asset generator) |
+| `range_proof` | bytes | ~3213 B (max 3328) | Borromean range proof |
+| `script` | bytes | variable (max 1024) | Locking script |
+| `asset_commitment` | bytes | 33 B | Blinded asset tag `A = H_token + r_asset*G` |
+| `surjection_proof` | bytes | variable (max 4096) | Asset surjection proof |
```
LFRezende, line 637: Why surjection proof is max 4096 B?

We can have at most 255 inputs (either transparent or not). 

For a given shielded output, we create a ring signature of `A_out - A_i`, for all `i=1, 2, ..., 255`.
The signature will have the initial challenge `e0` and the rest of the `s` factors, 32 B each.

```python

sigma = (e0, s1, s2, ..., s255)
```

Hence it should be at most `8192 B`, double of the mentioned. Are we limiting the input count (transparent or not) to 127?

I'm aware we capped the number of shielded elements.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+Hides the amount; token type remains visible.
+
+| Field | Type | Size | Description |
+|-------|------|------|-------------|
+| `commitment` | bytes | 33 B | Pedersen commitment `C = v*H_token + r*G` |
+| `range_proof` | bytes | ~3213 B (max 3328) | Borromean range proof |
+| `script` | bytes | variable (max 1024) | Locking script (P2PKH, etc.) |
+| `token_data` | int | 1 B | Token index (same semantics as `TxOutput.token_data`) |
+| `ephemeral_pubkey` | bytes \| None | 33 B | Compressed secp256k1 point for ECDH recovery (all-zero on the wire = not present) |
+
```
LFRezende, line 626: Again, this ephemeral key... It is related to Phase C, right?

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+Given:
+  Domain generators:   A_1, A_2, ..., A_n   (one per input + per MintHeader entry)
+  Output commitment:   A_out
+
+A ring proof over {A_out - A_i} proves knowledge of the discrete log for exactly
+one difference (the matching token), without revealing which.
+```
+
+The domain is built from: every transparent input's NUMS asset tag, every shielded input's on-chain `asset_commitment` (or derived tag for an `AmountShieldedOutput` input), and one NUMS asset tag per `MintHeader` entry (so a freshly-minted token can be the asset of a `FullShieldedOutput`). **Proof size**: grows with the domain size. Maximum: `MAX_SURJECTION_PROOF_SIZE = 4096` bytes.
+
```
LFRezende, line 579: Does the freshly (mint/melt)ed token tag get blinded as well? Or not?

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+```
+The committed amount v satisfies: 1 <= v < 2^40
+```
+
+(`htr-ct-crypto/src/rangeproof.rs`.)
+
+- **Lower bound `min_value = 1`** prevents zero-amount outputs (enforced both at proof creation and re-checked at verification).
+- **Fixed bit width `RANGE_PROOF_BITS = 40`.** 40 bits covers up to `2^40 − 1 ≈ 1.1 × 10^12` base units (~10 trillion HTR cents), above the maximum supply. Pinning the bit width makes every proof the **same size** regardless of the committed value, eliminating a proof-size side-channel.
+- **Proof size**: ~3213 bytes; bounded by `MAX_RANGE_PROOF_SIZE = 3328`.
+
```
LFRezende, line 558: Bit size is not 40. Still being decided but probably something between 64 and 160.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+Where:
+  H_token  = unblinded NUMS generator for this token
+  r_asset  = random asset blinding factor (32-byte scalar)
+  A        = blinded asset commitment (33 bytes, compressed point)
+```
+
+An observer sees `A` — a random-looking curve point indistinguishable from any other token's blinded commitment.
+
+### Borromean Range Proofs
+
```
LFRezende, line 546: We'll thankfully ditch it in favor of bulletproofs! :)

So this section might need to be deleted.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+### NUMS Asset Tag Derivation
+
+Each token has a deterministic generator `H_token` derived via a Nothing-Up-My-Sleeve (NUMS) construction (`htr-ct-crypto/src/generators.rs`):
+
+```
+tag = SHA256("Hathor_AssetTag_v1" || token_uid_32B)
+H_token = Generator::new_unblinded(tag)        # 33-byte point
+```
+
```
LFRezende, line 525: I may need to review the docs on `Generator::new_blinded()` to be sure, but the tag usually has a counter, like this:

```python
tag = SHA256("Hathor_AssetTag_v1" || token_uid_32B || k)
k = 0, 1, 2, ...
```

`H_token` is a hash-to-point procedure, so the hash MAY lie on the curve or not (~50% chance for secp256k1). 

In case it doesn't, we need to increase k and try again. The probability of it not lieing on a curve is `2**(-attemps)` so eventually it will fall. 

I pointed out in a previous repo review that the lib we used did not do this, but they seemed to use a different procedure. Nevertheless, it is worth flagging it out.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+```
+C = v * H + r * G
+
+Where:
+  v  = amount (u64)
+  r  = blinding factor (32-byte scalar)
+  H  = generator point specific to the token type
+  G  = secp256k1 base generator
+```
```
LFRezende, line 511: This holds only for ``AmountHidden` shielded outputs.

For `FullShielded` shielded outputs, `C = v*A + r_value*G`, where `A = H + r_asset*G` is the blinded asset tag. In amount only, `r_asset = 0`. 

We should have a small section disclaiming it, and I advise for us to handle mentally all Pedersen Commitments as `C == v*A + r*G`, to avoid mistakes in the future.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+| Show output amounts | Broken for shielded | Hidden behind Pedersen commitments |
+| Identify output token type | Broken for `FullShieldedOutput` | Hidden behind blinded asset commitments |
+| Compute address balances | Broken for shielded | Cannot sum opaque curve points |
+| Token supply tracking | **Unaffected** | Mint/melt amounts are declared publicly in `MintHeader`/`MeltHeader` |
+| Rich list / rankings | Broken for shielded | Balances not computable |
+
+### What Still Works
+
+- Transaction structure (input/output count, shielded vs. transparent).
+- Transparent outputs (amounts, tokens, scripts — unchanged).
```
LFRezende, line 318: This line depends on the choice we go with for the addresses. Some scripts may change (we may abandon P2PKH).

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+
+### Fee Computation
+
+```
+shielded_fee = (count_amount_shielded * FEE_PER_AMOUNT_SHIELDED_OUTPUT)
+             + (count_full_shielded   * FEE_PER_FULL_SHIELDED_OUTPUT)
+total_fee    = shielded_fee + standard_fee_if_any
+```
+
+For `DEPOSIT`-version tokens being minted/melted, the wallet must also fund (or will receive) the 1% HTR deposit/withdraw, which is folded into the balance equation rather than the `FeeHeader`. For `FEE`-version tokens, each `MintHeader`/`MeltHeader` entry costs one extra `FEE_PER_OUTPUT` (also folded into the balance equation). The wallet should display the expected fee before the user confirms.
```
LFRezende, line 301: Indeed it is fine to leave a 1% HTR deposit rate, as mint/melt amounts are public.

text/0000-shielded-outputs.md
```diff
@@ -0,0 +1,1120 @@
...
+7. **For `FullShieldedOutput`**: Create the blinded asset commitment and surjection proof, and embed `token_uid || asset_blinding_factor` in the range proof message.
+8. **Balance blinding factors**: Assign random blinding factors to all shielded outputs except the last, which receives the balancing residual computed by `compute_balancing_blinding_factor()`. (For full-unshield transactions there is no last shielded output, so the residual is published as the `UnshieldBalanceHeader` scalar.)
+9. **Compute and attach fee** in `FeeHeader`; attach `MintHeader`/`MeltHeader` if minting or melting.
+
+### Blinding Factor Management and Backup
+
+The wallet must store blinding factors for every owned shielded UTXO — without them, the funds cannot be spent.
+
+**Recovery guarantee**: Since blinding factors for received outputs are derived deterministically from `ECDH(scan_privkey, ephemeral_pubkey)` — and both the scan private key (derivable from the seed) and the ephemeral pubkey (stored on-chain) are always available — a **seed backup is sufficient** for full recovery. The wallet re-derives all blinding factors by scanning the blockchain and rewinding the range proofs.
```
LFRezende, line 275: Great comment. I like it.

LFRezende (commented):
Overall a really fine document, with only minor confusions and mistakes and/or just minor stuff that needs clarification. I think that after this first batch of reviews should already be approvable.

PR #105, opened 2026-03-05
rfc: OpenTelemetry observability for wallet-service

andreabadesso:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/11b236937b03fa6b2e7d2c840d7a451fc5c2f8b2/projects/wallet-service/0001-opentelemetry-observability.md)

## Summary

Adds distributed tracing, metrics, and structured logging to the wallet-service monorepo using OpenTelemetry.

**Key points:**
- Auto-instrumentation for mysql2, ioredis, aws-sdk, http — zero code changes for 80% of value
- Daemon: OTel SDK init + BatchSpanProcessor + K8s OTel Collector sidecar
- Lambdas: ADOT Lambda layer (managed by AWS)
- Backend: Grafana Tempo (traces, S3-backed) + Prometheus (span-derived metrics) — plugs into existing Grafana
- 6-phase rollout, each independently reversible

## Open questions

- ADOT cold start impact on our specific Lambdas (200-800ms range per AWS docs)
- Production sampling rates
- Trace context propagation through SQS between daemon and Lambdas


projects/wallet-service/0001-opentelemetry-observability.md
```diff
@@ -0,0 +1,473 @@
...
+
+## Sampling strategy
+
+For production, use a **tail-based sampling** strategy at the collector level:
+
+- Always keep traces with errors or high latency (>2s).
+- Sample 10% of successful, fast traces.
+- Always keep traces from critical paths (void handling, reorgs).
+
+This keeps storage costs manageable while ensuring we never miss problematic traces.
```
luislhl: We should also consider how long the traces will be kept in storage.

I believe we won't need to keep them for long, right? Only while we need to debug problems with them.

It seems there is a `compactor.block_retention`, which defaults to 14 days. We can use it and adjust if needed.

projects/wallet-service/0001-opentelemetry-observability.md
```diff
@@ -0,0 +1,473 @@
...
+- SQS message count — timeseries
+- Lambda-to-Lambda invoke duration — timeseries
+- Redis command duration (WebSocket session ops) — timeseries
+- External dependency error rate — timeseries
+
+**Data source:** Prometheus (span-derived metrics filtered by `span.name` matching `sqs.*`, `redis.*`, etc.)
+
+### Alerts
+
+All alerts use Grafana Alerting, evaluated against Prometheus metrics.
```
luislhl: We've been using Alertmanager alerts instead. We should keep the pattern unless you think there is a clear reason not to.

Alertmanager is what is integrated into our on-call setup currently.

They are defined in https://github.com/HathorNetwork/ops-tools/tree/master/kubernetes/infrastructure/kube-prometheus/alerts

projects/wallet-service/0001-opentelemetry-observability.md
```diff
@@ -0,0 +1,473 @@
...
+
+| Alert | Condition | Severity |
+|---|---|---|
+| API latency SLO breach | p99 > 5s for any endpoint over 5 min | Critical |
+| High error rate | Error rate > 5% for any endpoint over 5 min | Critical |
+| Slow DB queries | p95 query duration > 2s over 5 min | Warning |
+| Daemon sync lag | Sync lag > 100 blocks for 10 min | Critical |
+| Daemon event processing stuck | p99 processing duration > 60s over 5 min | Warning |
+| Lambda cold start regression | Avg cold start > 2s over 15 min | Info |
```
luislhl: The severities look wrong. I think we should use https://github.com/HathorNetwork/ops-tools/blob/master/docs/on-call/guide.md#alert-severitypriority as reference to define the severities.

It seems most of them would be Medium or less. Maybe the Daemon sync lag would be Major if we want to.

projects/wallet-service/0001-opentelemetry-observability.md
```diff
@@ -0,0 +1,473 @@
...
+
+All alerts link to the relevant dashboard panel for immediate drill-down. Critical alerts route to PagerDuty/Slack; warnings and info route to Slack only.
```
luislhl: We don't have PagerDuty, and don't need to worry routing alerts to Slack.

The alerting structure already does this for us, you just need to set alerts using Alertmanager as I said in other comment.

projects/wallet-service/0001-opentelemetry-observability.md
```diff
@@ -0,0 +1,473 @@
...
+
+- Always keep traces with errors or high latency (>2s).
+- Sample 10% of successful, fast traces.
+- Always keep traces from critical paths (void handling, reorgs).
+
+This keeps storage costs manageable while ensuring we never miss problematic traces.
+
+## Infrastructure changes
+
+Changes required in `ops-tools`:
```
luislhl, line 394: We will also need to create the S3 bucket and setup the permissions, preferably using Pod Workload Identity as is done in other Pods already. 

Since it's a side-container, it will probably use the IAM Role in the pod without any further config, but we may need to add permissions to it to access the new bucket: https://github.com/HathorNetwork/ops-tools/blob/master/terraform/AWS/kubernetes_iam_roles/hathor_wallet_service.tf

projects/wallet-service/0001-opentelemetry-observability.md
```diff
@@ -0,0 +1,473 @@
...
+
+Developers can view traces locally without deploying infrastructure by using the **console exporter** (prints spans to stdout) or running a local **Jaeger all-in-one** container:
+
+```bash
+docker run -d --name jaeger -p 16686:16686 -p 4318:4318 jaegertracing/all-in-one:latest
+```
+
+Set `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318` and traces appear at `http://localhost:16686`. This makes it easy to debug span hierarchy and timing during development.
+
+## Grafana dashboards
```
luislhl, line 308: There may be pre-built dashboards already, since we will be using standard metrics from OpenTelemetry. But we can also try to build them with AI to see what comes out.

projects/wallet-service/0001-opentelemetry-observability.md
```diff
@@ -0,0 +1,473 @@
...
+            L_AUTO["Auto-instrumentation:<br/>mysql, @aws-sdk/*, http, redis"]
+            L_ADOT --> L_AUTO
+        end
+    end
+
+    subgraph collector ["OTel Collector"]
+        OTLP_IN["OTLP Receiver"]
+        SMC["Span Metrics Connector<br/>(generates histograms & call counts)"]
+        OTLP_IN --> SMC
+    end
```
luislhl: This diagram apparently tells us that both the LAmbdas and the daemon will send traces to the same OTel collector.

But, as noted in some other parts of the text, for LAmbdas "the ADOT layer includes an embedded collector. Configuration is via a collector.yaml file bundled with the deployment, configured to export to the same Tempo endpoint."

So I think this diagram is a little misleading.

projects/wallet-service/0001-opentelemetry-observability.md
```diff
@@ -0,0 +1,473 @@
...
+    SMC -->|traces| TEMPO
+    SMC -->|span metrics| PROM
+
+    TEMPO -->|Tempo data source| GRAFANA
+    PROM -->|Prometheus data source| GRAFANA
+
+    GRAFANA -.->|"click latency spike<br/>→ jump to traces"| GRAFANA
+```
+
+Since we already run Grafana and Prometheus, the only new infrastructure component is **Grafana Tempo**. Tempo stores traces in S3 (cost-effective, no dedicated database) and integrates natively with Grafana as a data source. The OTel Collector's **Span Metrics Connector** automatically generates latency histograms and call count metrics from traces, which are scraped by our existing Prometheus. This gives us:
```
luislhl: I think we should add somewhere that Tempo will probably not need to be exposed to the Internet.

Both the Lambdas and Kubernetes live in the same VPC, so we probably can expose Tempo only internally in the VPC (like Grafana and Prometheus)

projects/wallet-service/0001-opentelemetry-observability.md
andreabadesso: Fixed. Split the diagram into two separate collector subgraphs: one for the K8s sidecar and one for the ADOT embedded collector in Lambda.

projects/wallet-service/0001-opentelemetry-observability.md
andreabadesso: Added a note that Tempo is VPC-internal only, same pattern as Grafana and Prometheus.

projects/wallet-service/0001-opentelemetry-observability.md
andreabadesso: Good point. Added a note to evaluate pre-built community dashboards before building from scratch.

projects/wallet-service/0001-opentelemetry-observability.md
andreabadesso: Switched to Alertmanager rules following the existing pattern in ops-tools/kubernetes/infrastructure/kube-prometheus/alerts.

projects/wallet-service/0001-opentelemetry-observability.md
andreabadesso: Fixed. Recalibrated using the on-call severity guide: sync lag as Major, API latency/error rate as Medium, the rest as Low.

projects/wallet-service/0001-opentelemetry-observability.md
andreabadesso: Removed the PagerDuty/Slack mention. Routing is now described as handled by the existing Alertmanager configuration.

projects/wallet-service/0001-opentelemetry-observability.md
andreabadesso: Added a note on trace retention: 14 days via compactor.block_retention (Tempo default), adjustable if needed.

projects/wallet-service/0001-opentelemetry-observability.md
andreabadesso: Added a new step 0 in infrastructure changes: create the S3 bucket and add permissions to the existing IAM role at terraform/AWS/kubernetes_iam_roles/hathor_wallet_service.tf via Pod Workload Identity.

luislhl (approved):

pedroferreira1 (approved):
For me this is approved, I just have some questions:

> Storage costs. Traces generate significant data volume. Without sampling, costs can grow quickly. Tail-based sampling mitigates this but adds collector complexity.

Do we have an estimation on how much this will increase?

> ADOT layer cold start impact in our specific Lambdas. The 200-800ms range is from AWS documentation. We need to benchmark with our actual deployment packages to get exact numbers.

We should implement this feature with a simple feature flag (even if it's an ENV var) to turn it off in a simple way, if we feel it's adding lot of overhead, not only to the cold start.

---

Would be amazing to have an analysis for each alert created for the alert manager on how to improve the query/API/error being alerted.

PR #111, opened 2026-05-06
design: initial shielded outputs support

r4mmer:
# Design: wallet-service shielded outputs support

Initial design proposal for indexing, registering, and serving **shielded outputs** in the `wallet-service`. The work is split into three sibling RFCs:

| Doc | Scope |
|---|---|
| [`0000-daemon-and-database.md`](projects/wallet-service/shielded-outputs/0000-daemon-and-database.md) | Schema, ingestion pipeline, reorg/void handling, push/WS event surfaces |
| [`0001-wallet-registration.md`](projects/wallet-service/shielded-outputs/0001-wallet-registration.md) | Key registration (creation + upgrade), HD derivation, historical re-scan |
| [`0002-shielded-outputs-wallet-api.md`](projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md) | Read API surface — balances, history, UTXOs |

## Motivation

Hathor is introducing **shielded outputs**, a new output type that hides amount (and optionally token) using Pedersen commitments, Bulletproofs, and surjection proofs. The `wallet-service` is the canonical indexer for hosted Hathor wallets and must index shielded outputs the same way it indexes transparent outputs.

The challenge is that shielded outputs are physically very different from transparent outputs: rows are an order of magnitude larger (~770–930 bytes of cryptographic material), ownership is determined by a different mechanism (script-extracted `spend_partial_address` rather than a fullnode-pre-decoded address), and recovering the value requires CPU-bound cryptographic work. Bolting these into the existing transparent code path would balloon row sizes, branch every hot-path query on a new dimension, and couple the well-tested transparent flow to an experimental new flow.

## Acceptance criteria

The design aims to satisfy the following goals:

- **Transparent path stays stable.** A bug in shielded handling cannot regress transparent balances, reads, reorg handling, or any existing endpoint.
- **Shielded ingestion mirrors transparent ingestion.** Every transparent capability — live ingest, reorg/void/unvoid, mempool, push/WS notifications — has a shielded counterpart with the same semantics.
- **Synchronous and consistent.** A vertex is either fully indexed (transparent + shielded) or not at all; no inconsistency window between the two views of the same tx.
- **Read path stays cheap.** No HD derivation on the API hot path; ownership matches are O(1) on an indexed key; unowned outputs cost essentially nothing.
- **Existing wallets can opt in.** A transparent-only wallet can attach shielded keys later and retroactively claim its historical receives via a bounded, idempotent, restartable catch-up scan.
- **API contract stays additive.** No `/v2`, no `Accept-Version` header. Old clients reading the same endpoints see a strict superset of today's response shape; existing fields keep their meaning.
- **Wallet-service never holds spend authority.** Only scan-side material (read access) is registered; spending stays a client responsibility.
- **Schema is forward-compatible.** Encryption-at-rest, nullifier-based consumption tracking, and shielded mint/melt support can all be added later without structural migrations.

## Design overview

The decisions below are the ones most likely to attract pushback. Each is argued in the corresponding RFC's *Rationale and alternatives* section; flagged here so reviewers can jump straight to the tradeoff.

### Separate tables, not a `kind` flag on `tx_output`

Six new tables (`shielded_tx_output`, `shielded_address`, `shielded_address_balance`, `shielded_wallet_balance`, `shielded_address_tx_history`, `shielded_wallet_tx_history`) hold all shielded state. Extending `tx_output` with nullable shielded columns and a `kind` discriminator was rejected, 99% of historical rows would carry columns they never use, every existing query would need a `WHERE kind` clause, and a bug in shielded ingestion could corrupt rows the transparent code path also reads. The only shared concept across the two paths is the **vertex**, not the storage layer.

### Synchronous in-line rewind on the daemon

When a shielded output's `spend_partial_address` matches an owned row, the daemon **synchronously** calls `rewindAmountShieldedOutput` / `rewindFullShieldedOutput` (~1 ms per match) inside the same per-vertex DB transaction. Async post-processing was rejected because it would open an inconsistency window between transparent and shielded views and complicate reorg ordering. The match step itself is O(1) on an indexed key, so unowned outputs cost essentially nothing.

### Per-index pre-derived scan privkeys

`shielded_address.scan_privkey` stores each child privkey ahead of time; the API never runs HD derivation on the read path. The wallet-service has two execution environments (daemon, Lambda API) and Lambda CPU per request is cost-sensitive, so HD derivation is moved to the gap-extension boundary which is rare.

### Wallet-load endpoint reused for shielded upgrade

There is **no** separate upgrade endpoint. The existing `wallet.load` handler (`POST wallet/init`) is extended with four optional fields (`scanXpriv`, `spendXpub`, plus signatures over both) and reconciles its stored state against the request. A reconciliation matrix in the RFC enumerates all six (wallet exists?) × (keys stored?) × (keys submitted?) cases, including a `409 Conflict` for the case where submitted keys differ from those already stored.

### Storage split, but the API merges by default

Storage stays per-kind (separate `tx_output` / `shielded_tx_output`, separate balance/history tables), but `GET wallet/balances` returns one merged number per token in the existing `balance.unlocked` field, and `GET wallet/utxos` returns merged results unless filtered with a new optional `kind=transparent|shielded` query parameter. A wallet UI almost always wants one number per token; doing the merge once in the wallet-service is cheaper than multiplying it across the client ecosystem. If a future feature needs the per-kind split exposed, the API can grow a query parameter additively.

### Backward-compatibility analysis

The RFC includes a per-endpoint table covering the four combinations of `{old client, new client} × {wallet has shielded keys, wallet doesn't}`. The conclusion: **no fund loss in any scenario**. The only observable old-client effects are in the rare *downgrade* scenario (user upgraded, then opened an old client), where they see a cosmetic balance discrepancy or stumble on a shielded UTXO at signing time — both recoverable.

### Out of scope

- **Encryption-at-rest** for scan-side secrets — schema is shaped to absorb it later.
- **Range-proof verification by the indexer** — currently delegated to the fullnode.
- **Wallet deletion / shielded-disable flow** — no transparent equivalent exists today.

## References

- Upstream parent RFC — [Shielded outputs (`feat/shielded-outputs`)](https://github.com/HathorNetwork/rfcs/tree/feat/shielded-outputs/text/0000-shielded-outputs.md)
- Upstream follow-on — [Shielded mint/melt RFC](https://github.com/HathorNetwork/rfcs/blob/feat/shielded-outputs/text/0000-shielded-outputs-mint-melt.md)
- Existing wallet-service repo — [`HathorNetwork/hathor-wallet-service`](https://github.com/HathorNetwork/hathor-wallet-service)


projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
```diff
@@ -0,0 +1,874 @@
...
+2. `handleVertexAccepted` builds the concatenated list and iterates:
+   - **`outputs[0]` (transparent, concatenated index 0).** INSERTs `tx_output` row with `mode = 0`, `address = Alice_transparent_address`, `value = 100`, `token_id = '00'`, `recovery_state = NULL`. The existing transparent ownership lookup against `address` (with `bip32_account = 0`) finds `wallet_id = wallet_alice`.
+   - **`shielded_outputs[0]` (mode=1, concatenated index 1).** INSERTs `tx_output` row with `mode = 1`, `address = Alice_spend_address`, `value = NULL`, `token_id = '00'` (from `token_data`), `recovery_state = 'unowned'`. INSERTs satellite row with crypto bytes. UPSERTs the `address` row for `Alice_spend_address` (already owned by `wallet_alice` at `bip32_account = 1`, `index = 7`). Rewinds → `value = 150`. UPDATEs the `tx_output` row to `value = 150`, `recovery_state = 'recovered'`.
+3. `updateAddressTablesWithTx` runs over both rows in the same call, dispatching on `mode`:
+   - `address_balance` for `(Alice_transparent_address, '00')`: bump `unlocked_balance += 100`, `total_received += 100`.
```
pedroferreira1: We don't have this separation anymore. Maybe this is just a reference that this is the address of the transparent output?

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
```diff
@@ -0,0 +1,874 @@
...
+
+```
+address
+  -- existing columns
+  address           VARCHAR(34)    NOT NULL,
+  wallet_id         VARCHAR(64)    NULL,
+  `index`           INT UNSIGNED   NULL,
+  transactions      INT UNSIGNED   NOT NULL DEFAULT 0,
+  -- new columns
+  bip32_account     TINYINT UNSIGNED NOT NULL DEFAULT 0,   -- 0 = transparent, 1 = shielded scan path
```
pedroferreira1: Maybe `0 = transparent` is not the best description. Isn't it better `0 = legacy`? Because the new addresses will be able to handle transparent outputs.

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
```diff
@@ -0,0 +1,874 @@
...
+- `total_received` / `total_shielded_received` are decremented here too, mirroring the transparent path's existing behavior — `totalAmountSent` is subtracted from `total_received` for the transparent contribution, and the analogous recovered-shielded receive amount is subtracted from `total_shielded_received` for the shielded contribution. Unrecovered shielded outputs (recovery_state ≠ 'recovered') never contributed to `total_shielded_received` and so don't decrement it on void.
+- `markUtxosAsVoided` is unchanged structurally — it works on `tx_output` rows uniformly, transparent and shielded.
+
+The shielded input case follows the same pattern: `prepareInputs` emits the input's contribution into `addressBalanceMap` reading `input.spent_output.mode` from the wire (see §*Spend / consume tracking*), so the helper sees a kind-tagged balance delta and dispatches columns accordingly.
+
+`handleUnvoidedTx` is the mirror — it calls `cleanupVoidedTx`, which today re-applies deltas with a forward sign; same shielded-column-dispatch extension applies.
+
```
pedroferreira1, line 753: We must add here the total_supply of tokens table update, in case a mint/melt tx gets voided.

projects/wallet-service/shielded-outputs/0001-wallet-registration.md
```diff
@@ -0,0 +1,489 @@
...
+
+1. Validates the existing transparent signatures (unchanged path).
+2. Validates `scanXprivSignature` and `spendXpubSignature` against `auth_xpubkey/0` to confirm the caller has authority over both new keys.
+3. Confirms the `wallet` row has no `scan_xpriv` stored (this is the upgrade case, not a key-mismatch case).
+4. Sets `wallet.scan_xpriv`, `wallet.spend_xpub`, `shielded_max_gap = 20`, `shielded_status = 'catching-up'`.
+5. Pre-derives 20 shielded addresses (indexes 0..19). For each, runs `INSERT … ON DUPLICATE KEY UPDATE` against the unified `address` table to either fill in ownership on a row the daemon had already observed, or insert a new owned row with `bip32_account = 1`. (`scan_pubkey` and `spend_pubkey` are computed during derivation but not stored.) New ownership rows are written with `catchup_state = 'pending'`.
+6. Enqueues a catch-up job for `wallet_alice`.
+7. Returns 200 with the wallet's status payload, including `shielded_status: 'catching-up'`. The client then polls the existing wallet-status endpoint until `shielded_status` reaches `'ready'`.
```
pedroferreira1: I'm thinking about this `shielded_status`. Maybe we can just have the `status` as we have now, otherwise the wallet library will need to handle both. If the wallet is 100% new, check status, if wallet is ready and doesn't have shielded keys (so it needs to load shielded addresses), change status to pending/loading, and return to the user. 

Does it make sense?

projects/wallet-service/shielded-outputs/0001-wallet-registration.md
```diff
@@ -0,0 +1,489 @@
...
+
+The Lambda environment makes per-request HD derivation a measurable cost; pre-derivation moves it to the gap-extension boundary which is rare. Storage cost of the per-index child privkeys is small.
+
+## Why not store the spend privkey too?
+
+The wallet-service must never hold the spend privkey. This is the entire point of the scan/spend split: the wallet-service can read but not steal. Spending is a client responsibility.
+
+## Why a separate `shielded_status` instead of reusing `wallet.status`?
+
+A wallet can be transparent-ready while shielded-catching-up. Conflating these into a single status enum forces the client to disambiguate via auxiliary fields. A separate column makes the lifecycle explicit.
```
pedroferreira1: I'm not sure I agree with it. The user wants to load the full wallet because he sent shielded keys, so we could re-use the `wallet.status` for it. I'm not sure I see benefits in having separate status. It doesn't harm but the code becomes more complex

projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md
```diff
@@ -0,0 +1,491 @@
...
+
+The core API design tension is between **backward compatibility** (old clients reading `balance.unlocked` as a single number expect a meaningful number) and **breakdown** (a privacy-aware UI wants to render transparent vs shielded explicitly). The chosen resolution: keep the existing field shapes as merged totals by default, and expose the breakdown via opt-in query parameters and new optional response fields. Old clients see numbers that match the user's spendable funds; new clients can read the breakdown without an extra request shape.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A developer building a wallet client thinks about three reads:
+
+1. **"How much HTR / token X does the user have?"** → balance endpoint. Gets back the user's complete balance per token (transparent + shielded merged). Pass `?include=split` to see the breakdown.
```
pedroferreira1: Does the `include` key accept other values?

In the headless wallet I added a `split=true` parameter to the balance API. What do you think?

projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md
```diff
@@ -0,0 +1,491 @@
...
+
+The core API design tension is between **backward compatibility** (old clients reading `balance.unlocked` as a single number expect a meaningful number) and **breakdown** (a privacy-aware UI wants to render transparent vs shielded explicitly). The chosen resolution: keep the existing field shapes as merged totals by default, and expose the breakdown via opt-in query parameters and new optional response fields. Old clients see numbers that match the user's spendable funds; new clients can read the breakdown without an extra request shape.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A developer building a wallet client thinks about three reads:
```
pedroferreira1, line 24: It's missing the address fetch. Would be good to have a new parameter to fetch legacy addresses. I'm using `legacy: false` in the new wallet clients. If `legacy == null` I return the old address.

projects/wallet-service/shielded-outputs/0001-wallet-registration.md
r4mmer: The issue starts when an old wallet connects to the wallet service, it will see loading/pending but will not know to send the new keys.

Having the client "opt-in" to shielded outputs makes it work with any wallet.

projects/wallet-service/shielded-outputs/0001-wallet-registration.md
r4mmer: Consider that once he sends the 2 sets of keys (account 0 and account 1) we need to start 2 processes to derive the keys, "adopt" the addresses to the wallet and manage the gap-limit.

These 2 processes can be parallel (for efficiency) and have 2 distinct status updates for each (issues, completed status, etc.).
Meaning that there can be a moment where process 0 (legacy) loads everything and process 2 is still loading.
This is the "transparent-ready while shielded-catching-up" referenced.

From my understanding we are handling 2 accounts in the same wallet and that both accounts have different rules, this can introduce some confusion if we try to conflate them so having some separation makes it easier to understand what is happening (in this case the state of the loading process)

projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md
r4mmer: I was thinking that later include could be a list so we can expand what information to fetch.

But I guess this was me overthinking it, having a boolean parameter is more than enough for the current needs.

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
r4mmer: Sorry, this is one of the cases where it is a bit confusing.

When a metadata change happens for a transaction where it becomes "unvoided" (it was voided=true, then updates to voided=false) the state machine triggers the `handlingUnvoidedTx` which calls `handleUnvoidedTx` and this method cleans up everything related to the transaction, then calls `handleVertexAccepted` on the same transaction.

It is assumed that all balance related to the unvoided tx was already removed when it became voided.
So the `handleVertexAccepted` will add everytthing back like it was the first time.

Which means that if we handle the mint/melt part of teh total_supply on the "new/void/reorg/remove" tx paths we can be sure they are up-to-date.

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
r4mmer: by "separation" you mean an address being a "transparent address"? or the columns on the database?

If it is the first one, I agree with you and will update the design

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
r4mmer: Good catch, will update the design with this

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
pedroferreira1: Yes but currently we handle the balance update using the output/input values. Now we can't trust these anymore because they may be shielded. So we need to explain how we are going to do with the mint/melt headers. That part that I missed in the RFC (we can add the mint/melt handling in another document, if you prefer)

projects/wallet-service/shielded-outputs/0001-wallet-registration.md
pedroferreira1: If an old client connects, we can continue with the current initialization that we have nowadays, nothing changes. If later the same wallet connects with a new client, we reset the initialization status.

What do you think?

pedroferreira1 (dismissed):

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
```diff
@@ -0,0 +1,910 @@
...
+2. Daemon's `handleVertexAccepted` runs. It iterates `outputs[]` (empty in this example) and then `shielded_outputs[]` (one entry).
+3. For the shielded entry, it INSERTs a `tx_output` row with:
+   - `mode = 1`
+   - `address = decoded.address` (base58 CTSpend-derived address)
+   - `index = 0` (concatenated index)
+   - `value = NULL`
+   - `token_id = '00'` (from `token_data`)
+   - `recovery_state = 'unowned'`
+4. It INSERTs a `shielded_tx_output_data` row with the crypto bytes.
+5. It UPSERTs `address`: row for Alice's `decoded.address` already exists with `bip32_account = Bip32Account.CTSpend`, `wallet_id = wallet_alice`, `index = 7`, `scan_privkey = …`, `ct_address = <long-form display string>` because Alice's wallet-registration step pre-derived this row earlier (registration is where the long-form is computed from `scan_privkey` + `spend_pubkey` and written to the row — see [[0001-wallet-registration]]). The single canonical `transactions` bump for `(address, tx)` happens once per vertex inside `updateAddressTablesWithTx` (see § *Invariants*); the observation upsert itself does not touch `transactions`.
```
tuliomir: fix(nitpick): Use GitHub markdown link syntax. `[[link]]` does not work in this interface.

```suggestion
5. It UPSERTs `address`: row for Alice's `decoded.address` already exists with `bip32_account = Bip32Account.CTSpend`, `wallet_id = wallet_alice`, `index = 7`, `scan_privkey = …`, `ct_address = <long-form display string>` because Alice's wallet-registration step pre-derived this row earlier (registration is where the long-form is computed from `scan_privkey` + `spend_pubkey` and written to the row — see [0001-wallet-registration](./0001-wallet-registration)). The single canonical `transactions` bump for `(address, tx)` happens once per vertex inside `updateAddressTablesWithTx` (see § *Invariants*); the observation upsert itself does not touch `transactions`.
```

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
```diff
@@ -0,0 +1,910 @@
...
+   - UPDATEs `tx_output`: `value = 75`, `token_id = 'T1'`, `recovery_state = 'recovered'`.
+2. `updateAddressTablesWithTx` updates `address_balance` for `(Alice_ctspend_address, 'T1')`: `unlocked_shielded_balance += 75`, `total_shielded_received += 75`.
+3. `wallet_balance` for `(wallet_alice, 'T1')`: `unlocked_shielded_balance += 75`. If this is the first time wallet_alice has seen token T1, the row is created by the upsert.
+4. `wallet_tx_history` row written with `shielded_balance_delta = 75` for `(wallet_alice, tx_id, 'T1')`. `address_tx_history` carries the same `shielded_balance_delta = 75` on the per-address row.
+
+The point of this example: `token_id` lives in `tx_output` and is filled in by recovery for mode 2, whereas for mode 1 it would have been filled at observe time. The handler dispatches on `mode` to know which.
+
+## Worked example: void of a recovered shielded receive
+
+Two days after the first worked example (Alice received 1.5 HTR shielded), the transaction is reorged out and gets voided.
```
tuliomir: suggestion(nitpick): What about _two minutes_? I know that theoretically it doesn't make a difference, but it's much closer to a concrete reorg we would actually see in the network.

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
```diff
@@ -0,0 +1,910 @@
...
+  script             VARBINARY(1024)     NOT NULL,  -- raw, not parsed
+  ephemeral_pubkey   VARBINARY(33)       NOT NULL,
+  token_data         TINYINT UNSIGNED    NULL,      -- non-null iff mode = 1; 1-byte wire value (low 7 bits = token index into vertex.tokens[], high bit = authority flag — always 0 for shielded)
+  asset_commitment   VARBINARY(33)       NULL,      -- non-null iff mode = 2
+  surjection_proof   VARBINARY(4096)     NULL,      -- non-null iff mode = 2
+  PRIMARY KEY (tx_id, `index`),
+  FOREIGN KEY (tx_id, `index`) REFERENCES tx_output(tx_id, `index`) ON DELETE CASCADE
+```
+
+The `index` column is `SMALLINT UNSIGNED` to match the widened `tx_output.index` (see § *Type-widening migration for `tx_output.index`*) and backticked because `index` is a SQL-reserved keyword in MySQL. Using the same column name and type as the parent table makes the FK self-documenting and removes the `output_index` ↔ `index` naming asymmetry an earlier revision had. `token_data` is `TINYINT UNSIGNED` to cover the full 0–255 byte range the wire emits — signed TINYINT (-128..127) would not fit `token_data` values ≥ 128.
```
tuliomir: question: Is it really safe to use a SQL-reserved keyword just to keep the FK self-documenting?

Would it be more maintainable to map the FK like the other tables/columns already have to do? I'm taking into consideration here that many LLM agents that will maintain this code may struggle with the idea that this very specific decision was taken for this table.

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
```diff
@@ -0,0 +1,910 @@
...
+async function handleVertexAccepted(db, txn, vertex) {
+  // 1. Block-specific bookkeeping (miner + heightlock). Existing logic.
+  if (isBlock(vertex)) await db.upsertMinerAndHeightlock(txn, vertex);
+
+  // 2. Add tx to transaction table. Existing logic.
+  await db.upsertTransaction(txn, vertex);
+
+  // 3-4. Single output loop spanning both kinds. Mark locked and insert all rows
+  // (transparent + shielded) into tx_output; shielded rows also write the satellite.
+  const concatenated = [
```
tuliomir, line 533: question: I believe this to be pacified between the Fullnode team and this RFC, but just to make sure:

The Fullnode will also provide outputs in this same order when asked about a transaction/output data, right?

Otherwise I can see a divergence in Fullnode/Wallet Service facade response behavior in the future spawning from this output handling such as what happened on https://github.com/HathorNetwork/hathor-wallet-lib/pull/1063 .

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
r4mmer: Updated to indicate that mint/melt with shielded outputs is going to be implemented after this design.

projects/wallet-service/shielded-outputs/0001-wallet-registration.md
r4mmer: Changed to merge the status columns into a single backwards compatible status

tuliomir (commented):
Partial review: only 0000 document

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
r4mmer: Sorry, there should not be a reference to old revisions.
The backticked index is already a part of the current tx_output table I am just using the same name on the new table.

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
r4mmer: I dont understand the question.
This is just a pseudo-code meant to ilustrate how the transaction will be processed.
The order of outputs from the api is not related to this part

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
r4mmer: Done

projects/wallet-service/shielded-outputs/0000-daemon-and-database.md
r4mmer: Done

projects/wallet-service/shielded-outputs/0001-wallet-registration.md
```diff
@@ -0,0 +1,497 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Schema additions on `wallet`
+
+New columns, all nullable to support the upgrade flow:
+
+```sql
+ALTER TABLE wallet RENAME COLUMN status TO legacy_status;  -- existing transparent lifecycle column
```
tuliomir, line 120: question(non-blocking): Should we really call transparent addresses/outputs as "legacy"? Is the intention really to stop using them as the main way of transmitting tokens, and favoring shielded outputs for everyone?

projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md
```diff
@@ -0,0 +1,558 @@
...
+### Response shape — merged list
+
+Every entry has a `kind` discriminator (`'transparent'` or `'shielded'`) so a client can dispatch on it. Common fields appear on both kinds; kind-specific fields appear only on the matching entries:
+
+| Field | Transparent | Shielded |
+|-------|-------------|----------|
+| `kind` | `'transparent'` | `'shielded'` |
+| `tx_id` | ✓ | ✓ |
+| `index` | ✓ (output index) | ✓ (concatenated index, see below) |
+| `token_id` | ✓ | ✓ (NULL for unrecovered `mode = 2`) |
```
tuliomir, line 297: thought: As much as we're trying to be backwards compatible, I think this has the potential to break some clients.

We'll certainly need a few integration tests to validate this on the Wallet Lib.

projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md
```diff
@@ -0,0 +1,558 @@
...
+      "voided":      false,
+      "version":     1,
+      "output_kind": "shielded",           // NEW. "transparent" | "shielded" | "mixed"
+      "balanceBreakdown": {                // NEW, optional
+        "transparent": 0,
+        "shielded":    150
+      }
+    },
+    {
+      "tx_id":       "cafebabe…",
```
tuliomir, line 120: You're getting even better with the hexadecimal funny names

deadbeef and cafebabe at a cafe

projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md
```diff
@@ -0,0 +1,558 @@
...
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+- **Server-side spend skeleton construction.** Once the spend wire-format is settled and a stable client library exists, the wallet-service could expose a `POST wallet/spend-shielded` endpoint (wallet identity from the bearer authorizer, mirroring the existing `tx/proposal` shape) that returns an unsigned transaction skeleton. The wallet-service still does not sign.
+- **Auditor-facing read endpoints.** With view-key delegation ([wallet registration](0001-wallet-registration.md) § Future possibilities), an auditor could read balances on a delegated read scope. The shape here trivially extends.
+- **GraphQL surface.** The merged/split pattern is a natural fit for GraphQL field-selection. Out of scope here, but worth keeping in mind if the team adopts GraphQL elsewhere.
+- **Push-notification payload extension** to include token symbol/name resolution server-side, sparing the client a follow-up call. Trivial follow-up.
```
tuliomir, line 557: question(non-blocking): Here you touch in interesting point that we'll only reach when this is deployed to mainnet. How will the push notifications behave when token_id is unknown?

Right now we dispatch the push right when the tx is received. With shielded outputs we would need to wait until the tx is fully decoded to know the amount and token uid (and hence the name and symbol) to send the user.

Additionally, this is only relevant if the user has requested the value on push. In case the user selected to only be informed of a new tx, the push can potentially be sent before the decryption happens.

tuliomir (approved):

projects/wallet-service/shielded-outputs/0001-wallet-registration.md
r4mmer: Yes, the initial version was transparent/shielded but the naming was changed to legacy since.

The idea that motivated this (from what was explained to me) is that a CT address has a spend and scan pubkeys in its composition, these can be used to create a shielded output, but the spend pubkey can also be used to create a simple P2PKH script to use in a transparent output.
So the new CT address can be used in both shielded and transparent outputs, making it a better address to use on wallets for both operations since the user does not have to track which address is for which type of operation.

projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md
r4mmer: I understand that having UTXOs without amount or token_data would break usual paths but remember that once the wallet has registered the CT pubkeys the wallet-service will return the values unshielded so the client will always receive UTXOs with known amount and token.

To spend and build these UTXOs will require special methods but usual paths on the wallet-service should work without much effort depending on the api changes that the fulnode facade changes may create.

projects/wallet-service/shielded-outputs/0002-shielded-outputs-wallet-api.md
r4mmer: Well, for the shielded outputs for the wallet receiving the notification we assume the CT pubkeys are registered already, so the CT address will be derived and the spend privkey to rewind the wallet's outputs on the transaction will be available.
This means we will be able to send the notification with the token and amount even for shielded outputs to the wallet.

A follow-up I realized now is that is multiple wallets are present on the trransaction the notification to wallet A should only have the rewound values of A's outputs and wallet B, while possible to rewind, we should send to A shielded (same for wallet B notification, keep A's outputs shielded)

pedroferreira1 (approved):

PR #113, opened 2026-05-14
docs(decimal-precision): add decimal precision analysis doc

glevco:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/master/projects/new-decimal-precision/0001-decimal-precision-analysis.md)


trondbjoroy (approved):
The technical tradeoff looks to be essentially free so we can base our decision on business and ecosystem needs. From that perspective, I agree with 18 decimals. 3 main reasons for that:

- EVM compatibility. Nearly all EVM tokens default to 18 decimals as it's the default in OpenZeppelin contracts. Bridging any of them to Hathor without a 1:1 decimal match would require truncation or rounding, introducing precision loss and potential trust issues as users expect exact amounts.

- Integration cost reduction. Developers bridging EVM assets or building cross-chain applications would need to handle decimal conversion logic if we used 8 decimals.

- Avoiding a second migration. Going to 8 decimals and later finding it insufficient for EVM would mean a second protocol change.

obiyankenobi (approved):

PR #115, opened 2026-06-02
feat(decimals): add token amount V2 RFC

glevco:
[Rendered](https://github.com/HathorNetwork/rfcs/blob/feat/decimal/token-amount-v2/projects/token-amount-v2/0002-token-amount-v2.md)


projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Custom tokens
+
+For fee-based tokens, the fee header will support V2 values (already covered in I5), and no further discussion is required. For deposit-based tokens, we must determine how the deposit and withdrawal of HTR will work. Currently, 1% of the minted value is required in HTR (rounded up) and the same 1% can be withdrawn when melting (rounded down). For example, minting 1.00 TOK requires 0.01 HTR, and minting 1.01 TOK requires 0.02 HTR. Melting 2.00 TOK yields 0.02 HTR, and melting 1.99 TOK yields 0.01 HTR. This means deposits and withdrawals always change in steps of 0.01 HTR, and therefore we decide:
+
+- **I8:** We keep the same step of 0.01 HTR for minting and melting custom deposit-based tokens, regardless of the increased precision. This keeps the economic model unchanged. For example, minting 1.001 TOK will require 0.02 HTR, and melting 1.999 TOK will yield 0.01 HTR. In other words, we apply the same 1% calculation, then round to the nearest HTR cent (up for deposits, down for withdrawals).
```
glevco, line 65: @obiyankenobi do you agree with this? Same discussion as in the project document.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Storage and indexes
+
+Storage of versioned transactions requires no upgrade, as they will naturally use their encoded bytes as the stored artifact. Token-related indexes need to be upgraded to support the new token amounts. This includes the UTXO index and the Tokens index.
+
+- **I9:** Indexes will use V2 values internally, as in I7. V1 values will be normalized accordingly.
+- **I10:** The RocksDB implementation of the UTXO index currently uses a fixed slot of 8 bytes for storing amounts. It will be changed to use a fixed slot of 16 bytes, which supports up to `~10^20` tokens with 18 decimal places.
```
glevco: Theoretically we may have token totals that are greater than this, in which case the index would break. We could use the varint encoding itself to encode the amount in the index, however I would need to check whether this would break any index operations. Another option is keeping a fixed size, just picking a greater one.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+- **I15:** V1 Blueprints can only be called by V1 transactions and other V1 Blueprints, and V2 Blueprints can only be called by V2 transactions and other V2 Blueprints. This ensures that all values in a single call chain are interpreted consistently, resulting in two separate Nano "worlds" that cannot interact directly. Runtime verification will be added to prevent cross-version calls (public, view, proxy) and balance retrieval.
+- **I16:** Upgrading a contract's Blueprint should work normally, and can be used to upgrade a contract from a V1 Blueprint to a V2 Blueprint, enabling support for the new decimal precision. Since the balance is normalized internally, this requires only correctly normalizing (or not) the values at the boundaries, according to the Blueprint's version (and gated by I15).
+- **I17:** We cannot drop support for V1 transactions while there are still V1 contracts in the network.
+- **I18:** The serialization of values and amounts in the Nano state trie must be updated to support V2 values. This will break the current state root ID.
```
glevco, line 100: Is that acceptable? Should we introduce a new Nano type, keeping the existing `Amount` as-is?

projects/token-amount-v2/0002-token-amount-v2.md
obiyankenobi: Yes, that is ok. In practice, people should not mint 0.000xxx. 

The only real discussion is about NFTs. Will we still consider 0.01 as a unit? Or should we migrate so people use fee-based tokens?

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+- **I1 (Implementation decision 1):** Use a 1-byte length-prefix varint (from the approved [analysis document](https://github.com/HathorNetwork/rfcs/pull/113)).
+- **I2:** The maximum amount of tokens per output will remain exactly the same but normalized to 18 decimals, that is, `2^63 * 10^16`, where `10^16` is hereby named the **normalization factor** — the factor used to scale V1 decimals to V2 decimals (from 2 to 18 decimal places).
+
+### Vertex versioning
+
+Existing vertices on the network form an immutable signed history with V1 token amounts that cannot be changed or migrated. We must add a versioning scheme that supports both V1 and V2 vertices concurrently, while still allowing them to interact with each other.
+
+- **I3:** Blocks do not need 18 decimal places for their rewards, so they will remain V1-only. This avoids the need to upgrade block templates and mining rigs.
+- **I4:** Transactions can be either V1 or V2, with the version indicated by a signal bit on the transaction (detailed in the [Reference-level section](#i4---versioning-transactions)). Support for V2 transactions will be gated by a Feature Activation process.
```
obiyankenobi, line 51: Would people still be able to send new V1 transactions? I think we should forbid it.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+### Wire encoding
+
+The current encoding of token amounts for the wire uses a custom varint of 4 or 8 bytes, with support for values of up to `2^63`, which includes the 2 decimal places. A new varint codec must be added to support the new range of amounts.
+
+- **I1 (Implementation decision 1):** Use a 1-byte length-prefix varint (from the approved [analysis document](https://github.com/HathorNetwork/rfcs/pull/113)).
+- **I2:** The maximum amount of tokens per output will remain exactly the same but normalized to 18 decimals, that is, `2^63 * 10^16`, where `10^16` is hereby named the **normalization factor** — the factor used to scale V1 decimals to V2 decimals (from 2 to 18 decimal places).
+
+### Vertex versioning
+
+Existing vertices on the network form an immutable signed history with V1 token amounts that cannot be changed or migrated. We must add a versioning scheme that supports both V1 and V2 vertices concurrently, while still allowing them to interact with each other.
```
obiyankenobi, line 48: Would that also mean a V2 for token creation?

projects/token-amount-v2/0002-token-amount-v2.md
glevco: > In practice, people should not mint 0.000xxx.

Why not? That's different from the PoC implementation. In it, the user can indeed mint `0.000xxx TOK`, and he'll pay `0.01 HTR` for it. If that's a rule and we'll forbid it, I have to add it.

> The only real discussion is about NFTs. Will we still consider 0.01 as a unit? Or should we migrate so people use fee-based tokens?

We can easily keep 0.01 as a unit for both versions, or we can use `10^-18` for V2. Then it's a matter of display only, in both cases we would show 1 according to the transaction's version. The latter would save a few bytes because 1 NFT would use 2 bytes, vs 4 in the current state.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Because V1 transactions will be necessary to interact with contracts that use V1 Blueprints. We can deprecate them after all contracts are migrated, and we invalidate those older Blueprints.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Yes. And for OCBs. But as explained, if we use the proposed approach which is using a signal bit, the token amount versioning is orthogonal to the transaction version, so it's a single change enabling support for all subtypes of transactions.

projects/token-amount-v2/0002-token-amount-v2.md
obiyankenobi: > Why not? That's different from the PoC implementation. In it, the user can indeed mint 0.000xxx TOK, and he'll pay 0.01 HTR for it. If that's a rule and we'll forbid it, I have to add it.

I said "in practice". I don't think people will mint 32.89893278347324387 USDT or some other token. They usually mint "rounder" numbers. But I agree it should be accepted by the protocol.

glevco: Even in practice, I'm not so sure. People are probably not going to _manually_ mint those numbers, but systems can easily do that, for example if they calculate amounts to mint based on some percentage.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: I changed it to use a varint, in https://github.com/HathorNetwork/rfcs/pull/115/changes/e08e5da11e57ead254112e2ac5d5a17c19c8318e

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+- **I3:** Blocks do not need 18 decimal places for their rewards, so they will remain V1-only. This avoids the need to upgrade block templates and mining rigs.
```
msbrogli, line 50: But miners would have to upgrade anyways. I guess we can migrate blocks too so everything is equal (but old vertices).

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: Should we put a minimum value for mint and melt operations?

msbrogli, line 65: I was wondering why not allow one to mint 0.0001 TKA depositing 0.000001 HTR. Any technical reason? Or just a business rule?

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+Storage of versioned transactions requires no upgrade, as they will naturally use their encoded bytes as the stored artifact. Token-related indexes need to be upgraded to support the new token amounts. This includes the UTXO index and the Tokens index.
+
+- **I9:** Indexes will use V2 values internally, as in I7. V1 values will be normalized accordingly.
+- **I10:** The RocksDB implementation of the UTXO index currently uses a fixed slot of 8 bytes for storing amounts. It will be changed to use a varint slot with the same serialization as the V2 token amounts.
+
+### APIs
+
+The full node APIs must reflect the new versioned tokens. For example, the `/transaction` route must return output values as they are defined in the respective transaction, so:
+
+- **I11:** V1 transactions will return V1-scaled token amounts in the `/transaction` API, and V2 transactions will return V2-scaled amounts. The API will also include a `token_amount_version` field so clients can correctly interpret the amounts considering the transaction's own version.
```
msbrogli: I guess we should create new fields and migrate all clients to use them. These new fields will provide values as strings, the number of decimals, and the version. v1 amounts should be scaled to 18 decimals just like v2 amount normally are. It might be a good time to create a v2 API.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+- **I12:** Existing APIs that take token amounts in the request will continue receiving V1 amounts to prevent breaking API contracts.
```
msbrogli: We should have a better migration plan here to prevent operations errors. I guess we should deprecate the v1 APIs and create a V2 APIs that will receive everything with 18 decimals.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+- **I13:** Existing APIs that return token amounts in the response will continue returning those amounts in the same field to prevent breaking API contracts; however, a new field with the same name suffixed with `_v2` will be added, containing the V2-scaled token amount.
```
msbrogli: Same here. Deprecate v1 and create a v2.

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: Maybe we should rename `Amount` to `2Amount` and create a new type `18Amount`. My name suggestions are horrible but I guess that's the best approach.  This would prevent blueprint devs from making mistakes assuming one thing or the other.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+Clients can therefore update their API calls incrementally while the full node is already using V2 accounting internally, at least until V2 transactions are activated (per I4). At that point, clients must be fully updated.
+
+### Nano contracts runtime
```
msbrogli, line 84: After Blueprint v2 is activated, will we allow the creation of v1 blueprints?

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+- **I16:** Upgrading a contract's Blueprint should work normally, and can be used to upgrade a contract from a V1 Blueprint to a V2 Blueprint, enabling support for the new decimal precision. Since the balance is normalized internally, this requires only correctly normalizing (or not) the values at the boundaries, according to the Blueprint's version (and gated by I15).
+- **I17:** We cannot drop support for V1 transactions while there are still V1 contracts in the network.
+- **I18:** The serialization of values and amounts in the Nano state trie must be updated to support V2 values. This will break the current state root ID.
+
+### Shielded outputs
+[shielded-outputs]: #shielded-outputs
+
+We must determine how the shielded outputs project will interact with this project.
+
+- Can we put values with increased precision in shielded outputs? Does that increase the required size of proofs?
```
msbrogli: Yes, it does increase the Borromean range proof since its size grows linearly.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+- **I15:** V1 Blueprints can only be called by V1 transactions and other V1 Blueprints, and V2 Blueprints can only be called by V2 transactions and other V2 Blueprints. This ensures that all values in a single call chain are interpreted consistently, resulting in two separate Nano "worlds" that cannot interact directly. Runtime verification will be added to prevent cross-version calls (public, view, proxy) and balance retrieval.
+- **I16:** Upgrading a contract's Blueprint should work normally, and can be used to upgrade a contract from a V1 Blueprint to a V2 Blueprint, enabling support for the new decimal precision. Since the balance is normalized internally, this requires only correctly normalizing (or not) the values at the boundaries, according to the Blueprint's version (and gated by I15).
+- **I17:** We cannot drop support for V1 transactions while there are still V1 contracts in the network.
+- **I18:** The serialization of values and amounts in the Nano state trie must be updated to support V2 values. This will break the current state root ID.
```
msbrogli, line 100: We cannot break the root ID. I'd say v1 contracts will keep storing with 2 decimals and v2 contracts will store with 18 decimals. This guarantees that all previous root ids remain the same.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Shielded outputs
+[shielded-outputs]: #shielded-outputs
```
msbrogli, line 103: We must ship this feature before shipping shielded outputs and enforce all shielded outputs to have 18 decimals. Otherwise, I guess we can't balance inputs and outputs with different versions.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+Option #2 is viable, but unnecessarily wasteful. Adding a new flag-only header would require 1 byte for the header version (with an empty body). All V2 transactions would have to carry that extra 1 byte, on top of the already increased encoded token amount sizes.
+
+Therefore, option #1 is the proposed approach: repurpose a signal bit in the transaction's first byte to encode the token amount version. The first byte of all vertices is a reserved signal byte intended precisely for this kind of upgrade. On mainnet/testnet, the least significant nibble of the signal byte of blocks is used for voting on Feature Activation processes. For transactions, the entire signal byte remains unused.
```
msbrogli: I guess the best way to put it is: Blocks and transactions share this byte during serialization. For blocks, this byte maps to signal bit field; for transactions, this byte should map to a separate flag field. I like it.

msbrogli (dismissed):

projects/token-amount-v2/0002-token-amount-v2.md
glevco: I guess it's just a business decision. Currently, it allows any value but rounds the deposit/withdraw to 1 cent of HTR. I like it because it's the most flexible for users, without changing the economical model.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: I don't see the advantage, only downsides. We do have to keep the V1 code forever, to support old vertices, so we don't have the advantage of removing code. And we also do have to support V1 transactions even after V2 is activated for interaction with existing Blueprints, so it's not like every new vertex in the network would be V2.

And the downsides:

- The current block reward of 2 HTR is encoded as 4 bytes in V1, and 9 bytes in V2.
- We would have to change the reward code / template generation to be dependent on Feature Activation. It currently uses V1 only. We've never used Feature Activation for changing block templates — although It should work seamlessly in theory, it's more code to change, review, and therefore more surface for bugs.
- What if external partners handle blocks using their own tooling? For example, miners generate output aggregators transactions. They wouldn't have to change anything if blocks keep being V1-only.

 It's increased block sizes and risks for no real gain.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Just a business rule, see https://github.com/HathorNetwork/rfcs/pull/115#discussion_r3414841463

projects/token-amount-v2/0002-token-amount-v2.md
glevco: I think we should, at least while there are contracts using _existing_ V1 blueprints  in the network, because it would be possible for an user to want to update his Blueprint to fix a bug for example, without wanting to change the logic for handling decimals in token amounts. If we forbid V1 blueprints, we would force them to change this logic which may not be easy.

After we're sure all existing contracts are migrated to V2 Blueprints, we can forbid V1 blueprints from being created, and more than that, all V1 transactions.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Then we have a problem of upgrading contracts from V1 to V2. By making the internal serialization the same across versions, upgrading would be seamless, just a matter of changing how the values are converted to and from the nano runtime.

If we can't do that, upgrading one contract would mean migrating what's stored at the moment of upgrade. That opens a new set of questions.

- Should we allow upgrades at all? If not, that's an important business decision. Contracts like Dozer would be locked in V1 forever.
- If we do, how do we migrate balances?

---

> We cannot break the root ID.

I guess an important question to make is _why_ this is true. Would that be the case even if we had tooling to make sure the migration was correct?

That relates to: what would be the plan if we ever found a determinism bug? It would mean different root IDs between nodes, so it would be a precedent for changing root IDs.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Updated in https://github.com/HathorNetwork/rfcs/pull/115/commits/ea9271811a9fcf0638c25a696ebeed4c0324bc4e 

What if a shielded tx uses non-shielded inputs from both V1 and V2? Would that be allowed?

projects/token-amount-v2/0002-token-amount-v2.md
glevco: I think this is just a matter of naming, "signal" is generic in this case and works as a synonym for "flag". All vertices have one signal byte (or one flag byte), and blocks use 4 bits for voting. 4 signal bits, 4 flag bits, reads the same for me.

So what do you mean by "for transactions, this byte should map to a separate flag field"?

projects/token-amount-v2/0002-token-amount-v2.md
glevco: I believe this depends on the discussion in https://github.com/HathorNetwork/rfcs/pull/115#discussion_r3416798099, so let's decide there first

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Updated in https://github.com/HathorNetwork/rfcs/pull/115/commits/ea9271811a9fcf0638c25a696ebeed4c0324bc4e

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Updated in https://github.com/HathorNetwork/rfcs/pull/115/commits/ea9271811a9fcf0638c25a696ebeed4c0324bc4e

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Updated in https://github.com/HathorNetwork/rfcs/pull/115/commits/ea9271811a9fcf0638c25a696ebeed4c0324bc4e

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Updated in https://github.com/HathorNetwork/rfcs/pull/115/commits/ea9271811a9fcf0638c25a696ebeed4c0324bc4e

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: Yes, but all shielded outputs must have 18 decimals. So v1 amounts in inputs or outputs would be scaled for balance check. The issue happens only on shielded amounts because the blinding factors make it impossible to multiply shielded amount by a constant (e.g., `10**16`). But we can easily do it on transparent amounts. It would be great to get this confirmed by @LFRezende .

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: After deserialization, I'd like to have it in another field (not `tx.signals`).

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: Actually, I guess we can. This value is not part of the block mining and it is used only to compare states between nodes. I guess all we need is a plan to prevent unnecessary warnings.

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: Thanks!

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: Thanks!

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: Thanks!

projects/token-amount-v2/0002-token-amount-v2.md
msbrogli: Thanks!

projects/token-amount-v2/0002-token-amount-v2.md
glevco: > Actually, I guess we can. This value is not part of the block mining and it is used only to compare states between nodes.

Exactly. Good, then!

> I guess all we need is a plan to prevent unnecessary warnings.

I don't think we currently have any automatic warnings triggered by this, only manual checks on releases.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: > So v1 amounts in inputs or outputs would be scaled for balance check

Perfect, that's already how it works.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Ok, got it! Added the name `flags_byte` to the reference-level on https://github.com/HathorNetwork/rfcs/pull/115/commits/433573d7bea6165a5e63213aed5a022caa1bd335

projects/token-amount-v2/0002-token-amount-v2.md
jansegre: I agree with Gabriel. I think it's just not worth making miners update their integration just to make block structures neater.

projects/token-amount-v2/0002-token-amount-v2.md
```diff
@@ -0,0 +1,264 @@
...
+
+The full node APIs must reflect the new versioned tokens. For example, the `/v1a/transaction` route must return output values as they are defined in the respective transaction, so:
+
+- **I11:** We'll introduce new `/v2` API routes for all existing `/v1a` routes that receive amounts as requests or send them as responses.
+- **I12:** The new `/v2` APIs will return amounts as strings, and include fields with the token amount version and number of decimal places.
+- **I13:** Existing `/v1a` APIs will be kept but deprecated, and they will return errors for V2 transactions.
+
+Clients can therefore update their API calls incrementally while the full node is already using V2 accounting internally, at least until V2 transactions are activated (per I4). At that point, clients must be fully updated.
+
+### Nano contracts runtime
```
msbrogli, line 84: What happens if a v2 contract calls a v1 contract? Does it fail? Or does it have to deposit and withdrawal amounts with 2 decimals only?

glevco: That's covered in I15:

> I15: V1 Blueprints can only be called by V1 transactions and other V1 Blueprints, and V2 Blueprints can only be called by V2 transactions and other V2 Blueprints. This ensures that all values in a single call chain are interpreted consistently, resulting in two separate Nano "worlds" that cannot interact directly. Runtime verification will be added to prevent cross-version calls (public, view, proxy) and balance retrieval.

glevco: In other words, that call is not allowed, it fails. That's implemented in PR https://github.com/HathorNetwork/hathor-core/pull/1736

msbrogli, line 84: I guess it's worth a more detailed description of the upgrade syscall when going from a v1 blueprint to a v2 blueprint (and vice versa, likely forbidden). Right after the syscall, the rest of the code will still run in a v1 environment, right?

projects/token-amount-v2/0002-token-amount-v2.md
obiyankenobi: I'm ok as Gabriel proposed. People can do whatever they prefer and we just round the numbers.

projects/token-amount-v2/0002-token-amount-v2.md
obiyankenobi: I guess it's better to force people to use V2 while the number of blueprints is low. In other words, do not allow the creation of V1 blueprints.

glevco: I agree, but I think we should enforce that on the OCB review instead of the protocol itself. So we allow existing contracts to be upgraded with bug fixes (an upgrade is a new Blueprint).

After all existing contracts are upgraded to V2, we can add this enforcement on the protocol.

projects/token-amount-v2/0002-token-amount-v2.md
glevco: Yes, I'll add that in the reference-level section that I'll write soon.

